PR #26, opened 2020-12-01
ci: surge preview

josokinas:
## Purpose

Deploy Storybook previews using [Surge](https://surge.sh/).

## Approach

Using [Surge PR Preview](https://github.com/marketplace/actions/surge-pr-preview) action deploying only for pull requests, removing preview when PR is closed (e.g. merged).

## Testing

Essentially when PRs are open at later stages, however tested [here](https://github.com/onfido/castor/pull/25).

## Risks

N/A

PR #32, opened 2020-12-02
ci: no surge teardown

josokinas:
## Purpose

Surge preview [was added](https://github.com/onfido/castor/pull/26) with possibility to teardown it after PR is merged/closed (following [official example](https://github.com/marketplace/actions/surge-pr-preview#teardown)).

After this now [every PR fails when merging](https://github.com/onfido/castor/pulls?q=is%3Apr+is%3Aclosed), not clear why but I suspect teardown does not work.

## Approach

Removing Surge preview teardown (at least for now).

## Testing

After merging PRs should not be failed.

## Risks

N/A

### DIFF
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -2,7 +2,6 @@ name: Preview
 
 on:
   pull_request_target:
-    types: [opened, synchronize, reopened, closed]
     branches-ignore:
       - 'release/*.*.x'
 
@@ -30,7 +29,6 @@ jobs:
             npx build-storybook -o public
           dist: 'public'
           failOnError: 'failed'
-          teardown: 'true'
 
       - name: Get preview URL
         run: echo "url => ${{ steps.preview_step.outputs.preview_url }}"


rabelloo:
Teardown seems to be working here https://github.com/iambumblehead/react-dropdown-now/blob/master/.github/workflows/preview.yml

Perhaps `on: pull_request` is the trick?

josokinas:
> Teardown seems to be working here https://github.com/iambumblehead/react-dropdown-now/blob/master/.github/workflows/preview.yml

Interesting. However, on this repo after merging I don't see any other pull requests, so can't be sure if that one does not fail.

But looking at their script is seems like there is no such line:

```yml
with:
  ref: refs/pull/${{ github.event.pull_request.number }}/merge
```

And then I checked [official example](https://github.com/marketplace/actions/surge-pr-preview#teardown) does not have this also... So probably this is what causing issues - will open a fix PR.

josokinas:
https://github.com/onfido/castor/pull/34 seemed to solved the issue, also looking at how https://github.com/onfido/castor/pull/35 was merged.

Closing this PR now.

josokinas:
Reopening as this is still an issue - will use like such until we _might_ switch to a different provider.

.github/workflows/preview.yml
```diff
@@ -2,7 +2,6 @@ name: Preview
 
 on:
   pull_request_target:
```
rabelloo, line 4: I believe this needs to be `pull_request` when no `types:` are specified. I'm not sure tho.

rabelloo (approved):

.github/workflows/preview.yml
josokinas: From what I [read on docs](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request_target) seems like they're same but safer to use secrets on?

rabelloo: Ah you're right, it needs `target` as it comments on the PR 👍🏻

PR #33, opened 2020-12-02
ci: correct naming main branch

josokinas:
## Purpose

Our primary branch is `main` not `master`, so it should also be used across CI scripts.

Incorrect name introduced with https://github.com/onfido/castor/pull/30.

## Approach

Rename primary branch as `main`.

## Testing

When merging to main branch deployment to GitHub Pages should start.

## Risks

N/A


rabelloo (approved):
😅

PR #36, opened 2020-12-03
feat(core): support 'raw' theme that doesn't include base tokens

rabelloo:
## Purpose

Enables efficiently including several themes by only emitting base tokens once.
When only 'day' and 'night' are used the difference is negligible.
However it scales much better for apps that require multiple themes.
The feature is opt-in, the default is still emitting all tokens.

## Approach

Refactor Sass theme mixins by combining `{theme}` and `{theme}-class`.
When `'class'` is specified as an argument, it will put all tokens inside the theme's appropriate class selector.
Otherwise it will use `&` which is essentially the same as not nesting inside a selector.

Then add the option of being `'raw'`.
When `'raw'` is specified as an argument, base tokens are not included in the output of the mixin.

## Testing

- Built styles (CSS) locally, output remains the same.
- Tested `'raw'` locally, does not output base tokens, only theme contextual/alias variables.

## Risks

BREAKING CHANGE: {theme}-class mixins removed in favour of {theme}('class')
See README.md changes for more details.


packages/core/README.md
```diff
@@ -112,8 +112,10 @@ You can also include class themes within your Sass file instead:
 ```scss
 @use '~@onfido/castor';
 
-@include castor.day-class();
-@include castor.night-class();
+:global {
```
josokinas, line 115: Original example was not using CSS modules, so I guess you either would need to remove `:global` or inform that this example is for when using CSS modules.

packages/core/README.md
```diff
@@ -132,6 +134,21 @@ switchTheme('custom', document.querySelector('.section'));
 
 If you do not use JavaScript, you might consider including a different CSS theme file based on URL parameter.
 
+Lastly, if you're extremely concerned about efficiency, you can shave off 1-3 KBs by not including base tokens twice, if they're shared between the themes you're switching.
```
josokinas, line 137: To follow other examples in this document:

```suggestion
Lastly, if you're extremely concerned about efficiency, you can shave off 1-3 KBs by not including base tokens twice, if they're shared between the themes you're switching:
```

packages/core/README.md
```diff
@@ -132,6 +134,21 @@ switchTheme('custom', document.querySelector('.section'));
...
+
+```scss
+@use '~@onfido/castor';
+
+:root {
+  @include castor.tokens();
+}
+
+:global {
```
josokinas, line 146: Here yet again using `:global` where as other examples are not using CSS modules.

packages/core/src/core/themes/day.scss
```diff
@@ -1,79 +1,84 @@
+@use 'sass:list';
 @use '../tokens' as castor;
 
-@mixin day() {
-  @include castor.tokens();
+// $type?: class | raw;
+@mixin day($type1: null, $type2: null) {
+  $class: list.index(($type1, $type2), 'class');
+  $raw: list.index(($type1, $type2), 'raw');
```
josokinas, line 7: [Minor] You might benefit a very little from `$types: ($type1, $type2);` 😏

```suggestion
  $types: ($type1, $type2);
  $class: list.index($types, 'class');
  $raw: list.index($types, 'raw');
```

packages/core/src/core/themes/day.scss
```diff
@@ -1,79 +1,84 @@
+@use 'sass:list';
 @use '../tokens' as castor;
 
-@mixin day() {
-  @include castor.tokens();
+// $type?: class | raw;
+@mixin day($type1: null, $type2: null) {
```
josokinas, line 5: Mixin seems to be repeated twice now, possibly more in the future if more themes to be added. Also, does not really help for whom is creating custom themes. Any possibility to have a build function that outputs the required mixin?

josokinas (changes requested):

josokinas:
I believe scope here should be `core` instead of `theme` as per [documentation](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#scope)?

packages/core/README.md
rabelloo: My bad here, I kind of interpreted this bit as a continuation of the previous topic. Will remove both instances of `:global`.

packages/core/README.md
rabelloo: Removed.

packages/core/src/core/themes/day.scss
rabelloo: Since it runs during build, not in runtime, doesn't matter much.
However, it does read better, I like that.

packages/core/src/core/themes/day.scss
rabelloo: I'll see what I can do.

rabelloo:
> I believe scope here should be `core` instead of `theme` as per [documentation](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#scope)?

Only if we choose to use package names as scope, like in [@commitlint/config-lerna-scopes](https://github.com/conventional-changelog/commitlint/tree/master/%40commitlint/config-lerna-scopes).

I don't mind doing that, even if not using `lerna-scopes`, so I'll amend the commit.

packages/core/src/core/themes/night.scss
```diff
@@ -1,79 +1,84 @@
+@use 'sass:list';
 @use '../tokens' as castor;
 
-@mixin night() {
-  @include castor.tokens();
+// $type?: class | raw;
+@mixin night($type1: null, $type2: null) {
+  $class: list.index(($type1, $type2), 'class');
+  $raw: list.index(($type1, $type2), 'raw');
```
josokinas, line 7: If you will not be refactoring these two theme mixins, then:

```suggestion
  $types: ($type1, $type2);
  $class: list.index($types, 'class');
  $raw: list.index($types, 'raw');
```

josokinas (approved):

packages/core/src/core/themes/night.scss
rabelloo: I probably will, just didn't have the time yet.
Can probably take another look today.
In the mean time I'll update to your suggestion.

packages/core/src/core/themes/day.scss
rabelloo: Unfortunately Sass doesn't support higher order functions (unless they're not anonymous, and it reads poorly).
However I've hoisted the theme logic to its own mixin, making it reusable across theme mixins.

It's still one line more than before, but overall not too bad.
Let me know what you think.

https://github.com/onfido/castor/commit/896eaca808ae5de1146e3651b4039b67e67d2414

josokinas: Looks great, merge! 😏

PR #53, opened 2020-12-07
refactor(core): folder structure

rabelloo:
## Purpose

Proposing a re-structure of our `core` folders.
Mainly wanted to get rid of `packages/core/src/core`.

## Approach

Moved files around and renamed imports where appropriate.

## Testing

GitHub actions. Manual.

## Risks

Totally subjective to personal choice, please let me know the bits you don't like and we can keep the good parts only.


josokinas:
Not sure about moving `mixins` to `utils` - is this intentional?

packages/core/src/utils/with-focus.scss
```diff
@@ -1,4 +1,4 @@
-@use '../core' as castor;
+@use '../helpers' as castor;
```
josokinas, line 1: Maybe better to use this a `helpers` (without aliasing to `castor`)?

packages/core/src/helpers/index.ts
```diff
@@ -2,3 +2,4 @@ export * from './border-radius/border-radius';
 export * from './color/color';
 export * from './font/font';
 export * from './space/space';
+export * from './switch-theme/switchTheme';
```
josokinas, line 5: Probably this file should be `switch-theme.js` to align with other helpers?

packages/core/src/themes/index.scss
```diff
@@ -0,0 +1,4 @@
+@forward './day';
+@forward './night';
+@forward './theme';
```
josokinas, line 3: If you agree with still having `mixins`, probably this should go there?

packages/core/src/utils/with-focus.scss
rabelloo: Fair point.

packages/core/src/helpers/index.ts
rabelloo: We have `utils/dashCase` and `utils/toCSS` though.
I guess this is as good a time as any to standardise file names.
Do you prefer `dash-case` or `camelCase`?
Should we match file name and its main export?

I'm actually leaning towards just matching folder/file/export name, thus `camelCase` makes more sense for TS.

josokinas: Originally idea was that:

- `helpers` are for mixed Sass/JS, that's why such naming.
- `mixins` for internal use, only Sass
- `utils` for internal use, only JS

Note JS is also TS here.

Now with this PR it's sort of mixed again... If you follow recomendation and split back to `utils` and `mixins`, then `helpers` could stay the same althought `switchTheme` is JS (TS) only.

packages/core/src/themes/index.scss
rabelloo: To answer your question, I moved `mixins` to `utils` intentionally, as I felt they have more or less the same goal.
I felt like `helpers` are things we export out, and `utils` are internal utilities.

I always prefer folder-by-feature than folder-by-type, but I suppose `@mixin theme` would fit in `utils`, by the logic above.

I'd rather not have `mixins`, but if you feel strongly about it, how's `utils/mixins`?

josokinas: I agree that they both same/similar but the file structure is different. For example, for utils we always have unit tests - for mixins we don't. That's why the `utils` directory is now a little unorganised IMO, it's more difficult to navigate.

Maybe mixins under utilities should also go into subdirectories?

packages/core/src/helpers/index.ts
rabelloo: I see. My issue with `mixins` is that we could potentially have `@function`s in there too 🤔 

I'll revert to split `mixins` and `utils` for now, as my main concern is the `core` folder.

Don't understand the bit about `switchTheme`, do you mean it should be in `utils`?
'Cause it's exported 🤔

josokinas: No, I mean `switchTheme` is only a JS helper, there's no Sass one. However, still I think matching naming as `switch-theme.js` is better for it?

Altough, why didn't you keep it under `themes`? I don't see it as a helper for core usage, like building the UI properties. Ahhh, this is why having `core` made sense, as it's a core items of Castor, rest being anything else like theme related.

Maybe simpliest would have been to rename `core` directory to `base` or alternative?

packages/core/src/themes/index.scss
rabelloo: I see. We should probably have unit tests for the mixins at some point.

I'll defer this issue for later, I just mean to refactor `core/src/core` after all.
For now I'll keep mixins.

packages/core/src/helpers/index.ts
rabelloo: I see. If we consider `helpers` as the core selectors, `switchTheme` really doesn't fit there.

Happy to move it under `themes`, let's see how that looks.

packages/core/src/utils/with-focus.scss
josokinas: My concern here is that previously `core -> castor` meant both helpers and other stuff e.g. tokens, themes. Now it's only helpers but still `castor`.

Now for the change, it then sort of makes sense - as users do import the whole Castor, but here internally it's only importing helpers.

packages/core/src/helpers/index.ts
josokinas: IMO looks better, however I'd personally name as:

```
/themes/switchTheme/
  - switchTheme.spec.ts
  - switchTheme.ts
```

packages/core/src/themes/index.scss
josokinas: I think utils should also be exported, but not meant to be used as an API. For example, someone might want to use `classy`.

packages/core/src/utils/with-focus.scss
rabelloo: Fair enough. Will change.

packages/core/src/helpers/index.ts
rabelloo: I prefer that too.
Now the only outlier is `border-radius` but I suppose it has a matching mixin in dash-case so we can keep it.

packages/core/src/themes/index.scss
rabelloo: Should they import from `@onfido/castor/utils` then?
Makes the official vs internal API explicit by path.

We might want to document that only the root export is the official API.

josokinas: Yes, it could be - looks nice, but previously you mentioned TypeScript issue about having imports from subdirectories?

josokinas (approved):
Looks good 👍

I wonder if choosing `helpers` is a good idea, with this slight confusion between "helpers" and "utils" maybe `selectors` would make more sense...

Also, still not 100% happy with `themes`. I'll leave this to you, but maybe this could do the trick:

```
/base/
  - /switchTheme/
  - /themes/
    - day.scss
    - night.scss
  - theme.scss
  - tokens.scss
```

packages/core/src/themes/index.scss
rabelloo: Can't remember what that was, but I don't foresee any issues.
Perhaps it was a matter of conflicting names?
We can revisit this problem if we encounter it in the future.

rabelloo:
I don't mind `selectors` vs `helpers`, even though we call them helpers in the docs.
I guess it's fine if the internal and external names are different as long as they don't leak boundaries.

Introducing a `base` folder would just re-introduce the `core` folder under a different name.
Ideally it'd be as flat as possible.

How do you feel about this:
```
/themes
  - day.scss
  - index.scss
  - night.scss
/theming
  /switchTheme
  - index.scss
  - index.ts
  - theme.scss
  - tokens.scss
```

josokinas:
> How do you feel about this...

That could work also, if your example demonstrates that `theme` mixin and tokens are on the root of `theming` but not inside `switchTheme`.

`themes` could also go inside `theming` - or it could be `theme/theming` and `theme/themes`.

rabelloo:
Changed, let me know how it looks now.

I prefer the flatness, but if you feel strongly I wouldn't oppose to e.g. :
```
/theme
  /themes
    - day.scss
    - index.scss
    - night.scss
  /switchTheme
    - switchTheme.ts
    - switchTheme.spec.ts
  - index.scss
  - index.ts
  - theme.scss
  - tokens.scss
```

josokinas:
> Changed, let me know how it looks now.

All good, just one thing - maybe you are better without `index.scss` within `themes`directory, and et the `index.scss` inside `theme` (or `theming`) do the job. You could still do `theme.day()` and `theme.night()`.

rabelloo:
Since there's no `themes/index.scss`, I felt like it belongs inside the folder where there is an `index.scss`.

How about now?

josokinas:
> How about now?

Nice, thanks for being patient with this.

Last thing, you need to adjust `src/index.scss` a little, and then good to go 🚀

rabelloo:
Of course, thanks for the thorough review 💪🏻 

Next up is `helpers/mixins/utils`, we can probably shape those up.
Will look at when I have some time.

Issue #1, opened 2020-11-20
Checkbox


rabelloo:
Implement the Checkbox component.

PR #78, opened 2020-12-10
feat: add checkbox component

josokinas:
## Purpose

Add Checkbox component.

## Approach

Implements Checkbox component.

This closes issue #1.

## Testing

On Storybook preview.

## Risks

N/A


packages/core/src/components/checkbox/checkbox.scss
```diff
@@ -0,0 +1,71 @@
...
+      width: $size;
+    }
+  }
+
+  // indicating [checked] or [indeterminate] input
+  .ods-checkbox:checked,
+  .ods-checkbox:indeterminate {
+    + .ods-input-indicator::after {
+      opacity: 1;
+    }
```
rabelloo: The `scale` here makes the "tick" and the "line" look blurry.
Also, we could use `transform-origin` rather than `margin` as it's more precise for translation.
Finally, there's an extra rule for `opacity`, but not necessary since the selector is the same.

Here's my suggestion.

| | before | after |
| --- | --- | --- |
| indeterminate | image | image |
| checked | image | image |

```suggestion    
    // "tick" for [checked] indication
    &:checked + .ods-input-indicator::after {
      $stroke-width: 2px;
      $width: $size * 0.65; // similar to `scale(0.6)`, but 0.65 makes it a nicer 14px
      $height: $width / 2;

      border: 0 solid helpers.color('content-on-action');
      border-width: 0 0 $stroke-width $stroke-width;
      height: $height;
      transform: rotate(-45deg);
      transform-origin: 110% 20%; // rotates rectangle from dead center
      width: $width;
    }

    // "line" for [indeterminate] indication
    &:indeterminate + .ods-input-indicator::after {
      $stroke-width: 2px;
      $width: $size * 0.55; // similar to `scale(0.6)`, but 0.55 makes it a nicer 12px
      $horizontal: ($size - $width) / 2;
      $vertical: ($size - $stroke-width) / 2; // vertical padding divided by two

      background-color: helpers.color('content-on-action');
      height: $stroke-width;
      margin: $vertical $horizontal;
      width: $width;
    }
  }
  
  // indicating [checked] or [indeterminate] input
  .ods-checkbox:checked,
  .ods-checkbox:indeterminate {
```

packages/react/src/components/checkbox/checkbox.react.stories.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+export const Playground: Story<CheckboxProps> = (props: CheckboxProps) => (
+  <Checkbox {...props} />
+);
+
+export const Bordered = storyOf(Checkbox, 'bordered', [true, false], {
+  labelMode: 'children',
+});
+Bordered.argTypes = omit<CheckboxProps>('bordered');
+Bordered.args = { name: 'bordered-playground' };
```
rabelloo, line 36: 💅🏻  [nitpick]
I don't think any of these stories need a `name`.
Especially since they're shared between each checkbox when they should be unique.

packages/react/src/components/checkbox/checkbox.react.stories.tsx
```diff
@@ -0,0 +1,81 @@
...
+  const checkboxRef = useRef<HTMLInputElement | null>(null);
+
+  const handleChange = ({
+    target: { checked: indeterminate },
+  }: ChangeEvent<HTMLInputElement>) => {
+    if (checkboxRef.current) checkboxRef.current.indeterminate = indeterminate;
+  };
+
+  return <Checkbox {...props} ref={checkboxRef} onChange={handleChange} />;
+};
```
rabelloo, line 60: I found it unintuitive that I had to click the checkbox to make the indeterminate state display.
I was expecting it to render in that state from the beginning.

Also, we are supposed to use these for visual regression testing later.

I'd recommend using `storyOf('indeterminate', [true, false])` instead.

packages/react/src/components/checkbox/checkbox.react.tsx
```diff
@@ -0,0 +1,36 @@
...
+      bordered,
+      invalid,
+      children,
+      className,
+      style,
+      ...restProps
+    }: CheckboxProps,
+    ref: CheckboxProps['ref']
+  ): JSX.Element => (
+    <InputContainer {...{ bordered, className, style }}>
```
rabelloo, line 18: 💅🏻  [nitpick]

Spreading props `{...{ bordered, className, style }}`, although terser, gives worse errors (TypeScript, ESLint).
For instance, if `bordered` had the wrong type, the component would have a red squiggly line instead of the offending property.

image

This is much clearer in my opinion:

image

rabelloo (changes requested):

packages/core/src/components/checkbox/checkbox.scss
josokinas: Thanks, applied [your suggestion but slitghly modified](https://github.com/onfido/castor/pull/78/commits/995f3a32e45a6d2c3ff0ea6704015585deac2ca3).

packages/react/src/components/checkbox/checkbox.react.stories.tsx
josokinas: I'm not sure about this.

We don't render stories for "checked" also.

Not sure "true" for such and then "false" would make sense, because you can still change the state, so when changed the label would not properly indicate the correct state. It could then disallow the state change, but also would be odd if interacting with an element does not do as it should.

That's why the story is called "as" - it does not demonstrate differences between one and another, but instead shows the state of such when it's enabled, that is when you "check". This same applies to Button component, that has a story "As Anchor".

packages/core/src/components/checkbox/checkbox.scss
rabelloo: Looks good, managed to get rid of magic numbers too 👍🏻

packages/core/src/components/checkbox/checkbox.scss
```diff
@@ -1,11 +1,11 @@
 @use '../../helpers';
+@use '../../internal';
 @use '../../mixins';
-@use '../input-container/input-container' as components;
 
 @mixin Checkbox() {
   $stroke-size: helpers.space(0.25);
 
-  @include components.InputContainer();
+  @include internal.IndicatorContainer();
```
rabelloo, line 8: We have to check whether including both Checkbox and Radio is going to repeat CSS from Indicator.

If it does, we probably shouldn't bundle, but instead document they're coupled

packages/react/src/components/checkbox/checkbox.react.stories.tsx
rabelloo: Maybe we should have `checked` stories?
It'll be clearer when we do have those visual tests.

The way I think about it is "`storiesOf` shows the default state when first rendered, but still allows interactivity".

Something we could do is controlled inputs, with a forced `checked` which means you can't change its state.
I wouldn't mind that if you prefer it.

josokinas: I actually disagree here...

O think `storyOf` should demonstrate prop variation.

`checked` and `indeterminate` is not part of props. Well, "checked" sort of is, if you decide to go the controlled way. But "indeterminate" is not - the code demonstrates only way this can be enabled via code.

But I do agree having state stories is valuable for visual tests, but maybe that's another bigger taks? For example, such would also be needed for Button's "pressed" state.

packages/core/src/components/checkbox/checkbox.scss
josokinas: Yes, it does compile the indicator container twice...

josokinas: What do you think, if internals would be included in components by default, but could also be excluded:

```scss
@mixin Component($include-internal: true) {
  @if $include-internal {
    @include internal.SomethingInternal();
  }
}
```

when want to compile all:

```scss
@mixin components() {
  $include-internal: false;

  @include castor.Internal();
  @include castor.Component($include-internal);
}
```

Docs should also clearly indicate such possibility, however then internals must be exported by default.

packages/react/src/components/checkbox/checkbox.react.stories.tsx
rabelloo: I see what you're saying, they're props to React but not specifically our component.
Still, we have different styles for each of those states.

Yea perhaps visual tests can be done for HTML + CSS, but it's also valuable to check that `<Checkbox indeterminate />` renders with the correct CSS.

Alright nevermind, this seems like a different and bigger discussion, let's leave as is for now.

josokinas: So yet `<Checkbox indeterminate />` will not work - you can't control this state like such...

packages/core/src/components/checkbox/checkbox.scss
rabelloo: That might be confusing in this very common scenario

```scss
:root {
  @include castor.Checkbox();
  @include castor.Radio($include-indicator: false);
}
```

This feels somewhat more natural, but also there's no telling they're coupled.

```scss
:root {
  @include castor.Checkbox();
  @include castor.Radio();
  @include castor.Indicator();
}
```

Another option is to include indicator in both, which allows them to be included separately.
Then when both are needed:

```scss
:root {
  @include castor.CheckboxAndRadio(); // comes with Indicator
}
```

Don't like that much though, doesn't scale well
`castor.CheckboxAndRadioAndInputAndTextarea()` 🤮 

Lastly, we could not export individual mixins and instead have logic in `components`:

```scss
@include castor.components((
  'button',
  'checkbox',
  'input',
  'radio',
));

@mixin components($list: ()) {
  @if 'Checkbox' in $list {
    @include castor.Checkbox();
  }

  @if 'Checkbox or 'Radio' in $list {
    @include castor.Indicator();
  }
}
```

The downside is loose strings instead of tight mixin names.

packages/react/src/components/checkbox/checkbox.react.stories.tsx
rabelloo: You're right, that to me is a bug though, React should control DOM properties when there's no attribute equivalent.

Apparently they are not supporting it because it can't be server-rendered https://github.com/facebook/react/issues/1798#issuecomment-333414857.

We should probably specify that prop and implement a `useEffect` hook to switch its `indeterminate` state.
Back to merging `ref`s fun.

As a Castor consumer I'd totally expect to be able to do `<Checkbox indeterminate />`

rabelloo (approved):

packages/core/src/components/checkbox/checkbox.scss
josokinas: I've changed this part now to track what additions are being included, so such are only included once per build.

packages/core/src/components/checkbox/checkbox.scss
```diff
@@ -0,0 +1,71 @@
+@use '../../helpers';
+@use '../../internal';
+@use '../../mixins';
+@use '../included';
+
+@mixin Checkbox() {
+  $stroke-size: helpers.space(0.25);
+
+  @if not included.is-included('indicator-container') {
```
rabelloo, line 9: 💅🏻  [nitpick]

I'd name it something cheeky but clearer like
```scss
@if not is.included('indicator-container');
```

```scss
@if not includes.the('indicator-container');
```

```scss
@if not already.included('indicator-container');
```

Or something less cheeky.

```scss
@if not internal.included('indicator-container');
```

I'd also just name it `indicator`, since it has styles for more than the container.

packages/core/src/components/included.scss
```diff
@@ -0,0 +1,15 @@
...
+  @if not map.get($_all-included, $key) {
+    $_current-included: (
+      $key: true,
+    );
+    $_all-included: map.merge($_all-included, $_current-included) !global;
+    @return false;
+  }
+
+  @return true;
+}
```
rabelloo, line 15: 💅🏻  [nitpick]

I'd change a few lines so that it's arguably clearer.

```suggestion
$_included: ();

@function is-included($key) {
  @if not map.get($_included, $key) {
    $_included: map.set($_included, $key, true) !global;
    @return false;
  }

  @return true;
}
```

rabelloo (approved):

packages/core/src/components/checkbox/checkbox.scss
josokinas: I don't want to have "internal" in the name, as I plan also to use the same for other cases, e.g. when one component requires another component (not internal one).

Also, would not want to go with just `indicator`, because internal component is called `IndicatorContainer` - I think better to match the name.

PR #80, opened 2020-12-10
chore: examples

### LINKED ISSUE #46: Add examples
Add examples of using Castor with different stack, e.g.:

- HTML + CSS
- HTML + CSS + JS
- React + Emotion
- React + Sass

etc.

Such would be in root directory `/examples` and would be part of Yarn workspaces.

Also, would be nice to deploy previews of each example when a pull request is opened.

josokinas:
## Purpose

Add examples of various Castor usage.

## Approach

Adding examples of:

- custom-themes
- html-css
- html-css-js
- html-css-js-parcel
- html-scss-ts-parcel
- react-emotion-ts-parcel
- react-scss-js-webpack
- react-scss-ts-parcel

Also adds Dependabot checks for each example, version gets also bumped on `package.json` of each.

Although discussed previously, I did not add previews of each example - this would make pull request pipeline to be very slow.

This closes #46.

## Testing

Manually locally.

## Risks

N/A


rabelloo:
> Although discussed previously, I did not add previews of each example - this would make pull request pipeline to be very slow.

Even if they run in parallel? What about running only for `main`?

josokinas:
> Even if they run in parallel? What about running only for `main`?

Still pipeline would block the pull request e.g. if merging until it finishes.

Maybe `main` would be better - just having those examples deployed somewhere.

.gitignore
```diff
@@ -21,5 +21,8 @@ package-lock.json
 *.tgz
 
 # Repository specific
-coverage
+.cache
+.castor*
+examples/*/dist
```
rabelloo, line 26: 💅🏻  [nitpick] I believe a simple `dist` covers any folder by that name anywhere.

rabelloo (approved):
I assume these are all copy pasta from `castor-examples`, except the root configs.
Hard to double check everything at once, but looks good.

rabelloo:
> Still pipeline would block the pull request e.g. if merging until it finishes.

If they ran as fast as our tests and linters, they'd never be the bottleneck.

But aside from just confirming the API didn't regress, don't see much value in previews for those, I suppose.
In which case we can simply build them, and not deploy.

josokinas:
> I assume these are all copy pasta from `castor-examples`, except the root configs.

It's definetly not a pasta 🍝 - but yes, I had to just do some slitgh adjustments due to linting configuration, then also did some renaming of examples to better identify what exactly each example consists of.

> But aside from just confirming the API didn't regress, don't see much value in previews for those, I suppose.
> In which case we can simply build them, and not deploy.

Cool, I'll do that then.

josokinas:
> What about running only for `main`?

I will leave this for now, I believe action we currently use is meant to only be use with pull requests.

PR #103, opened 2020-12-14
feat(core)!: add ods- prefix to all tokens

### LINKED ISSUE #102: Prefix tokens
We will have `ods` prefixed tokens.

rabelloo:
## Purpose

closes #102

## Approach

Search and replace.

## Testing

Local.

## Risks

Breaking change introduced.


josokinas:
There's a diff of alphabetical sort order in this PR. Would you mind to open a separate PR first, go through each file and save them so that it's all sorted and future code changes would not include it? It's just that it Git history would be cleaner that way...

Same goes to doc change - having it merged as a separate PR would be nicer 😏

rabelloo:
Sure thing, let me open separate PRs for those, then we can rebase this on top of that.

packages/core/src/theme/themes/day.scss
```diff
@@ -3,73 +3,75 @@
...
-    --color-background-inverse-action: var(--color-neutral-white), 1;
+    --ods-color-background-main: var(--ods-color-neutral-200), 1;
+    --ods-color-background-surface: var(--ods-color-neutral-050), 1;
+    --ods-color-background-surface-alt: var(--ods-color-neutral-300), 1;
+    --ods-color-background-action: var(--ods-color-primary-500), 1;
+    --ods-color-background-action-hover: var(--ods-color-primary-400), 1;
+    --ods-color-background-action-active: var(--ods-color-primary-600), 1;
+    --ods-color-background-action-subtle: var(--ods-color-primary-500), 0.1;
+    --ods-color-background-action-subtle-hover: var(--ods-color-primary-500),
+      0.05;
```
josokinas, line 34: How do you feel about now lines being over the limit, extending into 2nd line? I do like the limitation per config, however maybe worth increasing for token/theme files?

e.g. at the top of file:

```
/* stylelint-disable max-line-length */
```

josokinas (approved):

packages/core/src/theme/themes/day.scss
rabelloo: I don't much mind that, as other rules like complex `box-shadow`s usually span multiple lines too.

I wouldn't recommend disabling any lint rules, opens a precedent.
However we could change the max-line-length as a whole, or override for themes in specific.

josokinas: > However we could change the max-line-length as a whole...

ESlint has also a limit of 80 characters (if not mistaken), so IMO aligning these two to be the same is better...

rabelloo: Actually, tried `printWidth: 100`, only Prettier is needed, Stylelint doesn't complain.

rabelloo: I'd prefer to keep it at 80 (default), but if you want to change it, up to 90 is fine by me.
It's just enough to fit two files side-by-side in my editor and 24'' screen.

Just tried a few variations

| printWidth | files changed |
| -- | -- |
| 85 | 16 |
| 90 | 26 |
| 95 | 31 |
| 100 | 34 |

josokinas: I prefer to also stick to 80, as it's a standard. Happy for this to go now.

rabelloo: Oh, just realised something else, I don't think our GitHub actions for lint are picking up `warn`s 😱 

image

image

josokinas: 0 allowance of warnings is note set for Prettier, only for ESLint and stylelint. Same as on `package.json`.

rabelloo: 🤔  should it though?

josokinas: What's the parameter for Prettier? Don't see such [from available options](https://prettier.io/docs/en/cli.html).

rabelloo: [Prettier exits with 1](https://prettier.io/docs/en/cli.html#exit-codes) if any issue is found, I think the problem is that we're not including `scss` as an extension.

image

[The defaults are quite sensible](https://github.com/wearerequired/lint-action/blob/master/action.yml#L134), perhaps we should use them.

josokinas: Right, then probably [lint-staged config](https://github.com/onfido/castor/blob/main/package.json#L63) should also be adjusted.

rabelloo: Prettier and Stylelint actually conflict on these lines.
I've added line specific `// prettier-ignore` to them so that Stylelint can work its magic.
That good?

josokinas: Ignoring these tokens is fine, they still are very static entries.

However, maybe better to ignore the whole block, to avoid possible scenarios in the future?

Not entirely sure how this could be done, but most likely [it is possible somehow](https://prettier.io/docs/en/ignore.html), maybe:

```
// prettier-ignore-start
...
// prettier-ignore-end
```

Also, it states that `prettier-ignore`:

> will exclude the next node in the abstract syntax tree from formatting

So maybe you could try if adding at the top, or before curly brackets does the trick?

rabelloo: Having the comment just before the mixin works, it ignores the whole block.
However, I'd prefer to have the whole file linted, especially if it's auto-generated.

I see this as a temporary fix, we'll revisit this soon because of the Figma plugin and the cross-env token definitions you're working on.

I'm fine with a full block `// prettier-ignore` as a last resort.

josokinas: Then maybe `prettier-ignore-(start/end)` is better, surrounding only tokens nothing else in the file.

But your call.

Also, even with cross-env token definitions it might be that it will (somehow) end up looking the same. For example, it might be that it's auto-generated but commited as a file.

rabelloo: How do you feel about this? Saves 10 characters per line.

```scss
$prefix: --ods-color-;
$content: #{$prefix}content-;
$background: #{$prefix}background-;
$border: #{$prefix}border-;

#{$content}main: var(#{$prefix}neutral-800), 1;

#{$background}main: var(#{$prefix}neutral-200), 1;

#{$border}separator: var(#{$prefix}neutral-300), 1;
```

josokinas: Yes, fine. Thought should these be on `theme.scss` and reused across both themes?

Then I think it should be:

```scss
$prefix: --ods;
$color-prefix: #{$prefix}-color;
```

With that I also think that leading `-` should not be included in prefixes, but instead added when using, so:

```scss
#{$content}-main: var(#{$prefix}-neutral-800), 1;
```

What do you think?

rabelloo: The longer the variable, the less the savings. `$color-prefix` saves only 5 characters :/

The leading `-` is part of the prefix, as token names are simply `neutral-800`.
That's an important distinction for the Figma plugin.

If the issue is readability, there are other alternatives:

```scss
$ods: --ods-color-;
$content: $ods + content;
@function ods($name) {
     @return $ods + $name;
 }

#{$content}main: var($ods + neutral-800), 1;
// or
#{$content}main: var(ods(neutral-800)), 1;
```

I'd suggest either `$ods`, `$color` or `$prefix` for brevity and readability.
We could also do `$c` or `$p` but it loses meaning.

josokinas: I don't think we should care much about how many characters it saves - it's still processed to same CSS file, no matter which approach.

I would avoid variable name `$ods` as in case of a change of the prefix we use all variables would also need to be changed. Having more generic `$prefix` is better, IMO. But yet again, this would be for colors only, as for border radius (or anything else we might have) won't be able to be used.

rabelloo: The main reason for using those variables is to limit each line to 80 characters as not to exceed Prettier's printWidth rule 🤔

rabelloo: How about

```scss
@function color($name) {
  @return --#{$namespace}-color-#{$name};
}

$namespace: 'ods';
$content: color('content');
$background: color('background');
$border: color('border');

.foo {
  #{$content}-main: var(color(neutral-800)), 1;

  #{$background}-main: var(color(neutral-200)), 1;

  #{$border}-separator: var(color(neutral-300)), 1;
}
```

josokinas: Fine, as long as you won't mix them with `color` helper.

rabelloo: The helper needs the imported namespace, e.g. `helpers.color()`.

Do you want to merge as is with `prettier-ignore` and we look at this at a later stage?

josokinas: I would prefer to ignore the block, as I suggested, but as I already approved this - it's totally your call.

rabelloo: Cool, I'll do that.

PR #104, opened 2020-12-14
chore: lint import order

rabelloo:
## Purpose

Standardise import/export order.

## Approach

Use VSCode's "Organize Imports" feature on every file.

## Testing

Local.

## Risks

- Yet no automated CI way to enforce import/export order in the future;
- Since it auto-organises exports as well, some `.stories.tsx` would have their order changed, so I had to update code a bit to organised imports and arbitrarily sorted exports.

### DIFF
--- a/docs/_core/borders/borders.stories.tsx
+++ b/docs/_core/borders/borders.stories.tsx
@@ -1,5 +1,8 @@
+import { Primary, Stories, Title } from '@storybook/addon-docs/blocks';
 import React from 'react';
-import { Title, Primary, Stories } from '@storybook/addon-docs/blocks';
+import { Examples } from './examples.story';
+import { Helpers } from './helpers.story';
+import { Intro } from './intro.story';
 
 const docsPage = () => (
   <>
@@ -16,6 +19,4 @@ export default {
   },
 };
 
-export { Intro } from './intro.story';
-export { Examples } from './examples.story';
-export { Helpers } from './helpers.story';
+export { Intro, Examples, Helpers };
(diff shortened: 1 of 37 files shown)


docs/_core/borders/borders.stories.tsx
```diff
@@ -16,6 +19,4 @@ export default {
   },
 };
 
-export { Intro } from './intro.story';
-export { Examples } from './examples.story';
-export { Helpers } from './helpers.story';
+export { Intro, Examples, Helpers };
```
rabelloo, line 22: Here's an example of the refactor, had to, otherwise `export { Intro } from './intro.story';` would be last

examples/react-emotion-ts-parcel/src/index.html
```diff
@@ -1,8 +1,8 @@
-<!DOCTYPE html>
-<html lang="en">
-  <head>
-    <meta name="viewport" content="width=device-width,initial-scale=1" />
-    <title>React Emotion TS Parcel</title>
+<!DOCTY'@onfido/castor/dist/castor.css';
```
josokinas, line 1: That's probably a mistake?

josokinas (changes requested):

examples/react-emotion-ts-parcel/src/index.html
rabelloo: Yup, nice catch, I don't know how the build passed 🤔

josokinas (approved):

PR #109, opened 2020-12-16
feat: add search component

### LINKED ISSUE #61: Search
Implement the Search component.

rabelloo:
## Purpose

closes #61

## Approach

Implement Search styles and React component.

## Testing

Storybook.

## Risks

- Exposes Input ref
- Has (0, 2, 0) CSS specificity rules
- Has webkit specific style resets
- Props redirection is unorthodox - `className` and `style` are applied to container `<div>`, all the rest is applied to `<Input>`
- Icon requires the `<Icons>` sheet. Alternatively, inlining `<IconSearch />` would compromise efficiency (size)


packages/react/src/search/search.react.tsx
```diff
@@ -0,0 +1,20 @@
+import { c, classy } from '@onfido/castor';
+import { Icon } from '@onfido/castor-react';
+import React, { forwardRef } from 'react';
+import { Input, InputProps } from '../input/input.react';
+
+export const Search = forwardRef(_Search) as typeof _Search;
```
josokinas, line 6: What do you think if this is added:

```suggestion
/**
 * Please note that this component uses an Icon component that requires an SVG
 * sprite to be inlined in your app.
 * 
 * More details: https://github.com/onfido/castor-icons#use-with-plain-code
 */
export const Search = forwardRef(_Search) as typeof _Search;
```

packages/core/src/components/index.scss
```diff
@@ -4,12 +4,15 @@
...
+@use './search/search';
+@forward './search/search';
 @use './textarea/textarea';
 @forward './textarea/textarea';
 
 @mixin components() {
   @include button.Button();
   @include icon.Icon();
   @include input.Input();
+  @include search.Search();
```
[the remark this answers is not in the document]
rabelloo: You're right. Added now.
I wonder if we should still provide pre-built component CSS sheets 🤔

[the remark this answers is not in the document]
josokinas: Why not?

packages/core/src/components/search/search.scss
```diff
@@ -0,0 +1,31 @@
...
+@mixin Search() {
+  .ods-search {
+    align-items: center;
+    display: inline-flex;
+    position: relative;
+    width: helpers.space(50);
+
+    .ods-input {
+      border-radius: helpers.border-radius('full');
+      padding-left: helpers.space(2) + helpers.space(0.5);
```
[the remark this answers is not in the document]
rabelloo: I assume it's temporary, we'll wait for Steve's input.
In order to align text inside the Input with the full border-radius, 20px is needed for the left padding 🤷🏻 
If this is the value we end up going with, we should probably allow `space(2.5)`.

packages/core/src/components/search/search.scss
```diff
@@ -0,0 +1,31 @@
...
+      padding-right: helpers.space(6);
+      width: 100%;
+    }
+
+    .ods-input::-webkit-search-decoration,
+    .ods-input::-webkit-search-cancel-button,
+    .ods-input::-webkit-search-results-button,
+    .ods-input::-webkit-search-results-decoration {
+      display: none;
```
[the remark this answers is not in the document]
rabelloo: Added a comment, does this make sense?
```scss
// removes webkit specific X button/icon that clears the input
```

packages/react/src/input/input.react.tsx
```diff
@@ -1,17 +1,19 @@
...
-  className,
-  ...restProps
-}: InputProps): JSX.Element => (
-  <input
-    {...restProps}
-    type={type}
-    className={classy(c`input`, m({ invalid }), className)}
-  />
-);
+export const Input = forwardRef(_Input) as typeof _Input;
```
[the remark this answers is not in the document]
rabelloo: I'm really unsure. I have an actual use case for Search in specific, which is

```ts
useEffect(() => {
    if (!search) return;
    // focus and move cursor to end
    searchRef.current?.focus();
    searchRef.current?.setSelectionRange(-1, -1);
  }, [search]);
```

But it could be implemented in a different way, e.g.

```ts
export function Input({ focus }) {
  const ref = useRef<HTMLInputElement>();
  
  useEffect(() => {
    if (!focus) return;
    // focus and move cursor to end
    ref.current?.focus();
    ref.current?.setSelectionRange(-1, -1);
  }, [focus]);
  
  return <input ref={ref} />
}
```

I prefer limited APIs, but it's hard to preemptively cover all use cases, we'll probably need to keep tinkering.
Like, what if `blur` is now needed?

packages/core/src/components/index.scss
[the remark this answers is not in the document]
rabelloo: Diminishing returns, at some point it's better to have a bigger CSS sheet than several small ones, due to HTTP/1 limited amount of concurrent requests.

People that want to pick and choose components will use Sass, probably.

However, now I remember that with CSS imports you can also bundle these pre-compiled CSS sheets so there's value there, for people that don't want to use Sass, e.g.

```css
@import '@onfido/castor/dist/components/button';
@import '@onfido/castor/dist/components/icon';
```

packages/core/src/components/search/search.scss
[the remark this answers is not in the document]
josokinas: Right, well I assume the improved design will not have such border radius - looks way off other designs like input.

packages/react/src/input/input.react.tsx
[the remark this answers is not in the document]
josokinas: As mentioned some time earlier, I'd be pro having such by default on components.

packages/core/src/components/search/search.scss
```diff
@@ -0,0 +1,32 @@
...
+    width: helpers.space(50);
+
+    .ods-input {
+      border-radius: helpers.border-radius('full');
+      padding-left: helpers.space(2) + helpers.space(0.5);
+      padding-right: helpers.space(6);
+      width: 100%;
+    }
+
+    // removes webkit specific X button/icon that clears the input
```
josokinas, line 17:
```suggestion
    // removes WebKit specific "x" button/icon that clears the input
```

packages/core/src/components/search/search.scss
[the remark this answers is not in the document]
rabelloo: Yea probably so, I'll leave this conversation open so we don't forget to remove this code atrocity 😂

packages/react/src/input/input.react.tsx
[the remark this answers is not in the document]
rabelloo: Other than the horrific type juggling `const Foo = forwardRef(_Foo) as typeof _Foo;`, there's also this:

image

Which probably means we'll need `_Foo.displayName = 'Foo';` 😞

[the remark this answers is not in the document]
rabelloo: If you can find a nice solution for both problems, I'm happy to expose `ref`s for all base components (1:1 to an HTML element).

packages/core/src/components/search/search.scss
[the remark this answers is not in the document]
rabelloo: I've allowed `2.5` in `space` as it seems that's the correct design. Thoughts?

packages/react/src/input/input.react.tsx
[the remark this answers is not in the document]
rabelloo: I've create a helper `withRef` that "solves" both problems, please take a look and let me know what you think.

packages/react/src/utils/withRef.ts
```diff
@@ -0,0 +1,8 @@
+import { FC, forwardRef, ForwardRefRenderFunction } from 'react';
+
+/**
+ * Same as `forwardRef` except it returns the same type of the `component` going in.
+ * @param component Component to `forwardRef`.
```
josokinas, line 5: Cutting at 80 limit, adding space to align with all other fuction comments:

```suggestion
 * Same as `forwardRef` except it returns the same type of the `component` going
 * in.
 *
 * @param component Component to `forwardRef`.
```

packages/react/src/search/search.react.tsx
```diff
@@ -0,0 +1,22 @@
+import { c, classy } from '@onfido/castor';
+import { Icon } from '@onfido/castor-react';
+import React from 'react';
+import { Input, InputProps } from '../input/input.react';
+import { withRef } from '../utils';
+
+/**
+ * @note `Search` uses an `Icon` that requires
+ * `Icons` (SVG sprite) to be included in your app.
```
josokinas, line 9: To limit to 80 characters:

```suggestion
 * @note `Search` uses an `Icon` that requires `Icons` (SVG sprite) to be
 * included in your app.
```

Also, this comment style is different from the one on `Icon` component - mind amending that one too? Just to keep consistency.

packages/react/src/search/search.react.tsx
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * https://github.com/onfido/castor-icons#use-with-plain-code
+ */
+export const Search = withRef(
+  ({ className, style, ...props }: SearchProps, ref: SearchProps['ref']) => (
+    <div className={classy(c('search'), className)} style={style}>
+      <Input {...props} ref={ref} type="search" />
```
josokinas, line 16: To keep consistency across component code style:

```suggestion
  ({ className, style, ...restProps }: SearchProps, ref: SearchProps['ref']) => (
    <div className={classy(c('search'), className)} style={style}>
      <Input {...restProps} ref={ref} type="search" />
```

packages/react/src/search/search.react.stories.tsx
```diff
@@ -0,0 +1,27 @@
...
+  title: 'React/Search',
+  component: Search,
+  argTypes: {
+    ...omit<HTMLInputElement>('type'),
+    disabled: { table: { type: { summary: 'boolean' } } },
+    placeholder: { control: 'text' },
+  },
+  args: {
+    disabled: false,
+    placeholder: 'Placeholder',
```
josokinas, line 15: To align with `Input`:

```suggestion
    placeholder: { control: 'text' },
    maxLength: {
      control: 'number',
      table: { type: { summary: 'number' } },
    },
    disabled: {
      table: { type: { summary: 'boolean' } },
    },
  },
  args: {
    placeholder: 'Placeholder',
    disabled: false,
```

Otherwise `Input` story should be edited.

packages/react/src/search/search.react.stories.tsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import { Meta, omit, Story, storyOf } from '../../../../docs';
+import { Search, SearchProps } from './search.react';
+
+export default {
+  title: 'React/Search',
+  component: Search,
+  argTypes: {
+    ...omit<HTMLInputElement>('type'),
```
josokinas, line 9: Why not here:

```suggestion
    ...omit<SearchProps>('className', 'style'),
```

Curious 🤔

josokinas (changes requested):

packages/core/src/components/search/search.scss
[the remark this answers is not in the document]
josokinas: I don't like it... Seems like loosing purpose of 8px grid.

Maybe such cases better be hardcoded onto component, otherwise such now allos to use spacing in all apps.

josokinas:
Would we want to exclude type `search` from allowance for `Input` component?

packages/react/src/utils/index.ts
```diff
@@ -0,0 +1 @@
+export * from './withRef';
```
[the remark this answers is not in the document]
rabelloo: Gah I'd prefer not to. On the other hand it makes it look closer to `core`.

Would an `utils` folder outside of `src` be weird?
Perhaps we could rename `src` => `components` in that case.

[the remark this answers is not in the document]
josokinas: What's wrong with:

```
/src
  /components
  /utils
```

There might be other things in the future. And yes, it matches core more.

packages/react/src/utils/withRef.ts
rabelloo: That was 81 lines 🙈 
I've updated the comment.

packages/react/src/search/search.react.tsx
rabelloo: Ah I didn't notice Icon had a similar one, fixed.

packages/react/src/search/search.react.tsx
rabelloo: Do you prefer `restProps` or would you mind if I created another PR to rename them all to `props`?

josokinas: I prefer `restProps`. Reason for this that `props` come as an argument which here is destructed, and what's left is `restProps`. So `className + style + restProps = props`.

packages/react/src/input/input.react.tsx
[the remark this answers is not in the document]
josokinas: What's the plan, to forward all components with separate PR, or not really? Case by case?

packages/react/src/search/search.react.stories.tsx
rabelloo: I like the alphabetical table of props
image

But I'll keep them aligned and leave that for a future documentation update.

josokinas: I somehow don't think it will be possible to accomplish this when using the ones coming from type definition + custom table entries. But worth trying.

rabelloo: Oh but I'll omit `maxLength` from search, same as `invalid` is omitted, one without the other doesn't make much sense imo.

We should find a good way to document React standard props later.
Right now it's weird, e.g. if we include `maxLength` we should include `minLength`.

Ideally we could just link them and omit from the table.

packages/core/src/components/search/search.scss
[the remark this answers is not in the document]
rabelloo: To be fair I'm already using 20px in my team's product, so I'd benefit from it being allowed.

Another thing we didn't consider is negative values, like `space(-0.5)` isn't allowed but it probably should be.

I don't mind hard coding the value here, but it feels like we need to review `space`.

packages/react/src/search/search.react.stories.tsx
josokinas: When I added `maxLength` was thinking about `minLength` but it does not do anything. Probably even having `maxLength` is not worth it.

For example, try to sort out  `Button` component table arguments in alphabetical order...

packages/react/src/input/input.react.tsx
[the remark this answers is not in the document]
rabelloo: I think it makes sense to forward all form-related/input components, we can do that on a separate PR, yea.

For others I think case by case is better, like `Icon` probably doesn't need a ref?

packages/core/src/components/search/search.scss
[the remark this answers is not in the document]
josokinas: On the top of my head, probably `space` should not restrict on what multiplier is being passed? It should (probably) only do the calculation of base, then it's up to whoever is using to to follow the standard or not...

packages/react/src/utils/index.ts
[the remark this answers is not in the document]
rabelloo: I like flatter folder structures 😅 
But yea, I see the benefits here.

packages/react/src/input/input.react.tsx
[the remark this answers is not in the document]
josokinas: Not sure, we can start with form components yes, then see if there's a need for others. Thanks!

packages/react/src/utils/index.ts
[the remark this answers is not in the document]
josokinas: It's only one layer down, won't expect to have anything else 🙈

packages/react/src/search/search.react.stories.tsx
rabelloo: Well `omit<SearchProps>('className', 'style')` doesn't do anything, neither prop is included in the table.

Then, `type` is `Omit`ted from `InputProps` in `SearchProps`, but it was showing up in the table for some reason 🤔 
Just tried removing it and now it works, so I'll remove this line.

packages/react/src/search/search.react.stories.tsx
rabelloo: 🤔 
image

I can see your point tho, it can happen. In those cases we'd have to manually override them which is annoying

packages/core/src/components/search/search.scss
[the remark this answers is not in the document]
rabelloo: Yea I feel that way too, I don't like runtime checks anyway.
How do you feel about this thread @Subcide?

packages/react/src/search/search.react.stories.tsx
josokinas: So I somehow see limitation here with the plugin - perhaps we could submit a PR and add sorting option? What do you think, your Christmas project? 🎄

rabelloo:
> Would we want to exclude type search from allowance for Input component?

Because Search needs it, the workaround is annoying, we'd have to unsafe cast.

e.g.
```ts
export const Search = () =>
  <Input type={'search' as 'text'} />
```

Besides, people might want a `type="search"` input that differs in style from ours, so I think it's fine.

packages/react/src/search/search.react.stories.tsx
rabelloo: lol most definitely not, need some rest this holiday season.
But I can take a look when I'm back, sure.

packages/react/src/components/icon/icon.react.tsx
```diff
@@ -2,11 +2,11 @@ import { c, classy, color, IconProps as BaseProps } from '@onfido/castor';
 import React from 'react';
 
 /**
- * Please note that this component requires an SVG sprite to be inlined in your
- * app. If you wish to use individual inlined SVGs, use generated components
- * from Castor Icons package instead.
+ * @note `Icon` requires `Icons` (SVG sprite) to be included in your app.
+ * If you wish to use individual inlined SVGs, use generated components
+ * from `@onfido/castor-icons` instead.
```
josokinas, line 7: Reaching 80 limit:

```suggestion
 * If you wish to use individual inlined SVGs, use generated components from
 * `@onfido/castor-icons` instead.
```

josokinas (approved):

packages/core/src/components/search/search.scss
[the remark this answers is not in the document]
josokinas: Let's move this to a discussion at later stage - resolving this now.

packages/react/src/components/search/search.react.tsx
```diff
@@ -0,0 +1,25 @@
...
+    ref: SearchProps['ref']
+  ) => (
+    <div className={classy(c('search'), className)} style={style}>
+      <Input {...restProps} ref={ref} type="search" />
+      <Icon name="search" />
+    </div>
+  )
+);
+
+export type SearchProps = Omit<InputProps, 'children' | 'invalid' | 'type'>;
```
josokinas, line 25: Do you need `children` here?

rabelloo: Not anymore, since we're not including the label in `Search` anymore, so I omit `children`.
i.e.

```ts
<label>
  Foo bar
  <Search />
</label>
```

josokinas: I meant, do you need to omit it here? Does `InputProps` even have children?

rabelloo: It does, but nice catch, we should Omit there instead

rabelloo: Let me know how it looks now https://github.com/onfido/castor/pull/109/commits/8069044808b06209c1bbd0782cda76f20de0451e

josokinas: All good, just maybe `eslint-disable @typescript-eslint/no-explicit-any` could be on that one line only? Otherwise happy to be merged.

rabelloo: We need that in two lines.
Which of these do you prefer?

```ts
// eslint-disable-next-line @typescript-eslint/no-explicit-any
export const withRef = <C extends FC<any>>(component: C): C =>
  // eslint-disable-next-line @typescript-eslint/no-explicit-any
  forwardRef(component as any) as any;
```

```ts
/* eslint-disable @typescript-eslint/no-explicit-any */
export const withRef = <C extends FC<any>>(component: C): C =>
  forwardRef(component as any) as any;
/* eslint-enable @typescript-eslint/no-explicit-any */
```

```ts
// eslint-disable-next-line @typescript-eslint/no-explicit-any
export const withRef = <C extends FC<any>>(component: C): C =>
  forwardRef(component as never) as never;
```

josokinas: I do prefer 3rd option, however not sure I understand `never`? Is it better here than using `any`?

If it's not, then... Probably I would stick with 2nd option, just to have this bounded in a content insead of applied to the whole file.

rabelloo: They both achieve the same thing, though I'd argue `any` is clearer than `never`, the latter is misleading.

josokinas: So you choose whiever you think is best!

PR #111, opened 2020-12-16
refactor: use template literal types for terser code

rabelloo:
## Purpose

With newly introduced [template literal types](https://devblogs.microsoft.com/typescript/announcing-typescript-4-1/#template-literal-types) we can write terser code.

## Approach

Reuse type variants as much as possible in `color` and `font`.

## Testing

CI checks.

## Risks

Highly subjective to personal preference.

Note: the outcome is still the same, hovering the type name shows the same type union.

### DIFF
--- a/packages/core/src/helpers/color/color.ts
+++ b/packages/core/src/helpers/color/color.ts
@@ -38,88 +38,38 @@ export type Palette =
   | WarningColor
   | ErrorColor;
 
-type NeutralColor =
-  | 'neutral-900'
-  | 'neutral-800'
-  | 'neutral-700'
-  | 'neutral-600'
-  | 'neutral-500'
-  | 'neutral-400'
-  | 'neutral-300'
-  | 'neutral-200'
-  | 'neutral-050'
-  | 'neutral-white'
-  | 'neutral-black';
+type Weight =
+  | '900'
+  | '800'
+  | '700'
+  | '600'
+  | '500'
+  | '400'
+  | '300'
+  | '200'
+  | '100'
+  | '050';
 
-type PrimaryColor =
-  | 'primary-700'
-  | 'primary-600'
-  | 'primary-500'
-  | 'primary-400'
-  | 'primary-300'
-  | 'primary-200'
-  | 'primary-100'
-  | 'primary-050'
-  | 'primary-muted-600'
-  | 'primary-muted-300';
+type From700 = Exclude<Weight, '900' | '800'>;
+type From600 = Exclude<From700, '700'>;
 
-type AccentColor =
-  | 'accent-1-600'
-  | 'accent-1-500'
-  | 'accent-1-400'
-  | 'accent-2-600'
-  | 'accent-2-500'
-  | 'accent-2-400'
-  | 'accent-3-600'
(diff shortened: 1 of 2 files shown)


packages/core/src/helpers/font/font.ts
```diff
@@ -41,37 +41,19 @@ export function font(name: FontName): Font {
...
-  | '200-mono';
+  | `${FontSize}-bold`
+  | `${FontSize}-regular`
+  | `${LargeFont}-light`
+  | `${SmallFont}-allcaps`
+  | `${Exclude<SmallFont, '100'>}-mono`;
+
+type FontSize = keyof typeof sizes;
+type LargeFont = Exclude<FontSize, '400' | '300' | '200' | '100'>;
+type SmallFont = Exclude<FontSize, LargeFont>;
```
josokinas, line 56: At first glance this seems a little complex.

What do you think about such, even if it's longer:

```suggestion
  | `${FontSize}-bold`
  | `${FontSize}-regular`
  | `${Exclude<FontSize, '800' | '700' | '600' | '500'>}-light`
  | `${Exclude<FontSize, '400' | '300' | '200' | '100'>}-allcaps`
  | `${Exclude<FontSize, '400' | '300' | '200'>}-mono`;

type FontSize = keyof typeof sizes;
```

packages/core/src/helpers/color/color.ts
```diff
@@ -38,88 +38,38 @@ export type Palette =
...
-  | 'neutral-800'
-  | 'neutral-700'
-  | 'neutral-600'
-  | 'neutral-500'
-  | 'neutral-400'
-  | 'neutral-300'
-  | 'neutral-200'
-  | 'neutral-050'
-  | 'neutral-white'
-  | 'neutral-black';
```
josokinas, line 52: This all is much more difficult to read now...

What do you think if only like such would be a change:

```ts
type NeutralColor =   `neutral-${'900' | '800' | '700' | '600' | '500' | '400' | '300' | '200' | '050' | 'white' | 'black'}`;
```

Or use the `Weight` type, and exclude what's not needed for some cases?

josokinas (approved):

packages/core/src/helpers/color/color.ts
rabelloo: Fair enough. I reused types as most as I could, but it does make it slightly more annoying to read.

How does it look now? Want me to get rid of `Muted` and `Vivid` too?

rabelloo: Don't mind it, though I feel it makes it a little less readable, take a look.

josokinas: So I feel like such we now have only one place where available weight for colors is defined, and others only extract or exclude what they need.

rabelloo: Cool, unfortunately `Extract` does not suggest keys in the union so it's a bit iffy, but agree.
I'll merge like it is and we can always revisit it if we don't like it.

rabelloo:
Btw if you have ideas to use more of this feature in other places that'd be awesome.

PR #112, opened 2020-12-16
build: run linters for all appropriate files

rabelloo:
## Purpose

Make sure both GitHub Actions and lint-staged target all appropriate files for each linter.

## Approach

Keeping in mind that staged files already exclude files specified in `.gitignore`, in line with package.json scripts.

- Extract lint-staged config from package.json into its own config file (optional)
- Make `prettier` run for all staged files
- Make `eslint` run for all staged `js`, `jsx`, `ts` and `tsx` files
- Make `stylelint` run for all staged `css` and `scss` files
- Use `wearerequired/lint-action@v1` default extensions for `stylelint` and `prettier`, which cover all we need

## Testing

Update files of all linted extensions, commit, confirm all linters run correctly.

## Risks

lint-staged separate config file subject to personal taste.

### DIFF
--- a/.lintstagedrc.json
+++ b/.lintstagedrc.json
@@ -1,5 +1,5 @@
 {
-  "*": "prettier --write",
+  "*.{js,jsx,ts,tsx,css,scss,yml,html,md,mdx,json}": "prettier --write",
   "*.{j,t}s{,x}": "eslint --fix",
   "*.{,s}css": "stylelint --fix"
 }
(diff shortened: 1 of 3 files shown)


.lintstagedrc.json
```diff
@@ -0,0 +1,5 @@
+{
+  "*": "prettier --write",
```
josokinas, line 2: Not sure with which file type, but I was having issue with such, can you try locally a few files like `.gitignore` or `.nvmrc` (stage them).

josokinas (approved):

.lintstagedrc.json
rabelloo: You're right, that's unfortunate :/
Fixed. Wish we could efficiently reuse our scripts instead, will investigate in the future.

josokinas: Nothing is stopping us from just using them on the workflow, we don't need to use that action... We only loose some nice UI on PR.

rabelloo: Yea, I mean I'll see if there's a way for us to get the best of both worlds.

PR #113, opened 2020-12-16
refactor(core): simplify classy

### LINKED ISSUE #47: Simplify `classy` utility
Currently `classy` is very complex only due to support of various usage.

After discussion with @rabelloo we have decided that template literal support is not needed for such util and can be removed in order to simplify and make `classy` smaller.

rabelloo:
## Purpose

Reduce maintenance cost at little to no code ergonomic impact.
closes #47 

## Approach

- Cut support for template strings which vastly simplifies `c` and `m`
- Reuse some logic between `classy` and `withPrefix`, namely `truthyNames`

## Testing

- Jest unit tests
- Build process

## Risks

`c` and `m` no longer support template strings

### DIFF
--- a/packages/core/src/utils/classy/classy.ts
+++ b/packages/core/src/utils/classy/classy.ts
@@ -29,7 +29,7 @@ export const classy = (...classNames: ClassNames[]) =>
  * c('button');
  * // ['ods-button']
  */
-export const c = mapWithPrefix('ods-');
+export const c = withPrefix('ods-');
 
 /**
  * Prefixes modifier names with a predetermined identifier.
@@ -40,7 +40,7 @@ export const c = mapWithPrefix('ods-');
  * m('action', 'primary');
  * // ['-action', '-primary']
  */
-export const m = mapWithPrefix('-');
+export const m = withPrefix('-');
 
 const truthyNames = (value: ClassName | ClassNameRecord) =>
   (value instanceof Object
@@ -50,7 +50,7 @@ const truthyNames = (value: ClassName | ClassNameRecord) =>
     : [value]
   ).filter(Boolean);
 
-function mapWithPrefix(prefix: string) {
+function withPrefix(prefix: string) {
   return (...classNames: (ClassName | ClassNameRecord)[]) =>
     classNames.flatMap(truthyNames).map((name) => `${prefix}${name}`);
 }
--- a/packages/core/src/utils/classy/classy.ts
+++ b/packages/core/src/utils/classy/classy.ts
@@ -1,7 +1,6 @@
 export type ClassName = string | null | undefined;
 export type ClassNameRecord = Record<string, boolean | null | undefined>;
 export type ClassNames = ClassName | ClassName[] | ClassNameRecord;
-type Template = [TemplateStringsArray, ...ClassNames[]];
 
 /**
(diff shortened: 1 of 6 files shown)


josokinas:
Probably we should treat it like a breaking change?

packages/core/src/utils/classy/classy.ts
```diff
@@ -19,73 +18,39 @@ type Template = [TemplateStringsArray, ...ClassNames[]];
...
-    .flat(10)
-    .flatMap((className) =>
-      className instanceof Object ? getTruthyKeys(className) : className
-    )
-    .filter(Boolean)
-    .join(' ');
+  classNames.flat(10).flatMap(truthyNames).join(' ');
 
 /**
  * Prefixes component names with a predetermined identifier.
```
josokinas, line 24: I suppose it should be singular?

```suggestion
 * Prefixes component name with a predetermined identifier.
```

packages/core/src/utils/classy/classy.ts
```diff
@@ -19,73 +18,39 @@ type Template = [TemplateStringsArray, ...ClassNames[]];
...
+
+const truthyNames = (value: ClassName | ClassNameRecord) =>
+  (value instanceof Object
+    ? Object.entries(value)
+        .filter(([, value]) => !!value)
+        .map(([key]) => key)
+    : [value]
+  ).filter(Boolean);
+
+function withPrefix(prefix: string) {
```
josokinas, line 53: Curious, why did you choose `truthyNames` to be an arrow function, and this a regular one?

josokinas (approved):
Code looks much cleaner now, thanks!

packages/core/src/utils/classy/classy.ts
rabelloo: Well theoretically it could be multiple, e.g. `c('card', 'accordion')` 🤷🏻

packages/core/src/utils/classy/classy.ts
rabelloo: I prefer arrow functions in general, except when the function has a body (multiple expressions).

However this one didn't leave me much choice as it's used in lines 32 and 43 so it needs to be hoisted.

packages/core/src/utils/classy/classy.ts
josokinas: Ah, yes, you're right! Ignore this...

rabelloo:
> Probably we should treat it like a breaking change?

I wouldn't, because it's an internal function, not part of the external/official API.

packages/core/src/utils/classy/classy.ts
josokinas: I guess something to think about - maybe we would want to standartise this, e.g. use one kind only.

josokinas:
Please mention that this closes #102.

packages/core/src/utils/classy/classy.ts
rabelloo: If we were to, I suppose they'd all need to be regular functions, because sometimes we need the hoisting.

We can have guidelines for sure, extending eslint's rules.
There are many instances where arrow functions are much terser.

rabelloo:
> Please mention that this closes #47.

Done. Forgot about the issue.

PR #116, opened 2020-12-17
refactor(core): make "space" helper similar between .ts and .scss

rabelloo:
## Purpose

Refactors `space` for better maintainability, error messages, variable reutilisation, and parity between TS and Sass.

## Approach

- Reuse `space`'s `allowedFloats` to interpolate floating numbers allowed
- Rename variables to match between languages
- Reorganise code structure to make them better match

## Testing

`npm run test`

## Risks

None?

### DIFF
--- a/packages/core/src/helpers/color/color.ts
+++ b/packages/core/src/helpers/color/color.ts
@@ -22,9 +22,8 @@ export function color(name: Color): string;
 export function color(name: Palette, opacity?: number): string;
 
 export function color(name: Color, opacity?: number): string {
-  return opacity
-    ? `rgba(var(--ods-color-${name}), ${opacity})`
-    : `rgba(var(--ods-color-${name}))`;
+  const alpha = opacity ? `, ${opacity}` : '';
+  return `rgba(var(--ods-color-${name})${alpha})`;
 }
 
 export type Color = ContentColor | BackgroundColor | BorderColor | Palette;
--- a/packages/core/src/helpers/color/color.ts
+++ b/packages/core/src/helpers/color/color.ts
@@ -22,8 +22,9 @@ export function color(name: Color): string;
 export function color(name: Palette, opacity?: number): string;
 
 export function color(name: Color, opacity?: number): string {
-  const alpha = opacity ? `, ${opacity}` : '';
-  return `rgba(var(--ods-color-${name})${alpha})`;
+  return opacity
+    ? `rgba(var(--ods-color-${name}), ${opacity})`
+    : `rgba(var(--ods-color-${name}))`;
 }
 
 export type Color = ContentColor | BackgroundColor | BorderColor | Palette;
--- a/packages/core/src/helpers/space/space.scss
+++ b/packages/core/src/helpers/space/space.scss
@@ -3,7 +3,7 @@ $base: 8px;
 
 @function space($multiplier) {
   @if not is-allowed($multiplier) {
-    @error '$number must be an integer or one of (#{$allowed-floats})';
+    @error '$multiplier must be an integer or (#{$allowed-floats}). Got: #{$multiplier}';
   }
 
   @return $multiplier * $base;
--- a/packages/core/src/helpers/space/space.scss
+++ b/packages/core/src/helpers/space/space.scss
@@ -1,11 +1,18 @@
-$allowed-fractionals: (1.5, 0.5, 0.25, 0.125);
-
+$allowed-floats: (1.5, 0.5, 0.25, 0.125);
 $base: 8px;
 
-@function space($number) {
-  @if $number % 1 != 0 and not index($allowed-fractionals, $number) {
-    @error '$number must be an integer or one of (#{$allowed-fractionals})';
+@function space($multiplier) {
+  @if not is-allowed($multiplier) {
+    @error '$number must be an integer or one of (#{$allowed-floats})';
   }
 
-  @return $number * $base;
+  @return $multiplier * $base;
+}
+
+@function is-allowed($multiplier) {
+  @return is-integer($multiplier) or index($allowed-floats, $multiplier);
+}
+
+@function is-integer($number) {
+  @return $number % 1 == 0;
 }
--- a/packages/core/src/helpers/space/space.scss
+++ b/packages/core/src/helpers/space/space.scss
@@ -3,7 +3,7 @@ $base: 8px;
 
 @function space($multiplier) {
   @if not is-allowed($multiplier) {
-    @error '$multiplier must be an integer or (#{$allowed-floats}). Got: #{$multiplier}';
+    @error '$multiplier must be an integer or (#{$allowed-floats})';
   }
 
   @return $multiplier * $base;
--- a/packages/core/src/helpers/space/space.ts
+++ b/packages/core/src/helpers/space/space.ts
@@ -1,7 +1,7 @@
 /**
  * Returns the `px` size of a multiplier of the base space.
  *
- * Only integers and (1.5, 0.5, 0.25, 0.125) are supported.
+ * @note Only integers and (1.5, 0.5, 0.25, 0.125) are supported.
  *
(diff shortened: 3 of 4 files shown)


packages/core/src/helpers/space/space.ts
```diff
@@ -10,16 +10,17 @@
...
  */
 export function space(multiplier: number): string {
-  if (!isAllowed(multiplier)) {
-    const msg = '"multiplier" must be an integer or (1.5, 0.5, 0.25, 0.125)';
-    throw new Error(`${msg}. Got: ${multiplier}`);
-  }
+  if (!isAllowed(multiplier))
+    throw new Error(
+      `"multiplier" must be an integer or (${floats}). Got: ${multiplier}`
+    );
```
josokinas, line 16: I like that allowed fractionals are reused, however not sure `floats` constant is needed - won't be reused anywhere. So what if:

```suggestion
    throw new Error(
      `"multiplier" must be an integer or (${[...allowedFractionals].join(', ')}). Got: ${multiplier}`
    );
```

packages/core/src/helpers/color/color.ts
```diff
@@ -22,9 +22,8 @@ export function color(name: Color): string;
 export function color(name: Palette, opacity?: number): string;
 
 export function color(name: Color, opacity?: number): string {
-  return opacity
-    ? `rgba(var(--ods-color-${name}), ${opacity})`
-    : `rgba(var(--ods-color-${name}))`;
+  const alpha = opacity ? `, ${opacity}` : '';
+  return `rgba(var(--ods-color-${name})${alpha})`;
```
josokinas, line 26: I must say... it maybe cutting some size, but decrease readability.

josokinas (approved):

packages/core/src/helpers/space/space.ts
rabelloo: That was my first approach, the problem is prettier 🙊 
```
throw new Error(
    `"multiplier" must be an integer or (${[...allowedFractionals].join(
      ', '
  )}). Got: ${multiplier}`
);
```

packages/core/src/helpers/color/color.ts
rabelloo: Yea... Would a helper make it more legible?

```ts
export function color(name: Color, opacity?: number): string {
  return `rgba(var(--ods-color-${name})${maybe(opacity)})`;
}

const maybe = (value: unknown) => (value ? `, ${value}` : '');
```

Or can you think of a different way?

It might be too early to optimise for size anyway.
If you're not happy with any approach, I don't mind reverting this.

josokinas: With helper looks little better. But anyhow I feel like original implementation matches Sass one better.

packages/core/src/helpers/space/space.ts
josokinas: Not sure, but I would like this to match Sass function as close as possible.

```ts
export function space(multiplier: number): string {
  if (!isAllowed(multiplier))
    throw new Error(
      `"multiplier" must be an integer or one of (${[
        ...allowedFractionals,
      ].join(', ')}). Got: ${multiplier}`
    );

  return `${multiplier * base}px`;
}
```

I also think that Sass function should also use `multiplier`, so:

```scss
@function space($multiplier) {
  @if $multiplier % 1 != 0 and not index($allowed-fractionals, $multiplier) {
    @error '$multiplier must be an integer or one of (#{$allowed-fractionals})';
  }

  @return $multiplier * $base;
}
```

packages/core/src/helpers/color/color.ts
rabelloo: We could refactor the one in Sass to be about the same, but I agree on the readability aspect.
Sass doesn't benefit from cutting implementation details size as it doesn't ship to the browser.

I'll just revert then, we can worry about that in the future, probably won't be our bottleneck anyway.
BTW we should probably include bundle size checks in our CI, like [bundlesize](https://github.com/siddharthkp/bundlesize) or [bundlewatch](https://github.com/bundlewatch/bundlewatch)

josokinas: Yeah, cool - please could you [add an issue](https://github.com/onfido/castor/issues).

packages/core/src/helpers/space/space.ts
rabelloo: Fair enough, take a look at the new version and let me know how that looks.

packages/core/src/helpers/space/space.ts
```diff
@@ -1,7 +1,7 @@
 /**
  * Returns the `px` size of a multiplier of the base space.
  *
- * Only integers and (1.5, 0.5, 0.25, 0.125) are supported.
+ * @note Only integers and (1.5, 0.5, 0.25, 0.125) are supported.
```
josokinas, line 4: Why `@note`? I feel like such should be introduced per whole codebase with separate PR...

packages/core/src/helpers/color/color.ts
rabelloo: https://github.com/onfido/castor/issues/118

packages/core/src/helpers/space/space.scss
```diff
@@ -1,11 +1,18 @@
...
-
+$allowed-floats: (1.5, 0.5, 0.25, 0.125);
 $base: 8px;
 
-@function space($number) {
-  @if $number % 1 != 0 and not index($allowed-fractionals, $number) {
-    @error '$number must be an integer or one of (#{$allowed-fractionals})';
+@function space($multiplier) {
+  @if not is-allowed($multiplier) {
+    @error '$number must be an integer or one of (#{$allowed-floats})';
```
josokinas, line 6: Either:

```suggestion
    @error '$multiplier must be an integer or (#{$allowed-floats})';
```

or

```suggestion
    @error '$multiplier must be an integer or (#{$allowed-floats}). Got: #{$multiplier}';
```

josokinas (changes requested):

packages/core/src/helpers/space/space.ts
rabelloo: Fair enough, removing.

packages/core/src/helpers/space/space.scss
rabelloo: Ah missed that, used editor name refactor feature, thanks!

packages/core/src/helpers/space/space.scss
```diff
@@ -1,11 +1,18 @@
...
-
+$allowed-floats: (1.5, 0.5, 0.25, 0.125);
 $base: 8px;
 
-@function space($number) {
-  @if $number % 1 != 0 and not index($allowed-fractionals, $number) {
-    @error '$number must be an integer or one of (#{$allowed-fractionals})';
+@function space($multiplier) {
+  @if not is-allowed($multiplier) {
+    @error '$multiplier must be an integer or (#{$allowed-floats}). Got: #{$multiplier}';
```
josokinas, line 6: Only now this exceeds 80 characters...

I leave this up to you, but maybe we don't need `Got: <multiplier>` on any...

josokinas (approved):

josokinas:
P.S. If you could also amend the PR description...

packages/core/src/helpers/space/space.scss
rabelloo: Yea... there's the stack for that. I'll remove the last bit.

rabelloo:
One last review of the `spec` refactor?

josokinas:
> review of the `spec` refactor?

Personally I'm not a fan of such testing - where it seems like the test itself is an implementation. I'd rather have hardcoded tests with input out. But your call, already approved.

rabelloo:
Fair enough, reverted.

josokinas:
> Fair enough, reverted.

Sorry 🙈

PR #117, opened 2020-12-18
refactor: rename dashcase to kebabcase

josokinas:
## Purpose

Looking at other libraries, such function is to be [called `kebabCase`](https://lodash.com/docs/4.17.15#kebabCase).

## Approach

Proposal is to rename `dashCase` utility to `kebabCase`.

## Testing

Unit tests.

## Risks

Breaking change for those who would use the utility, however API is not exposed so this should only be for internal use.


rabelloo (approved):
Personally I dislike the name, but [I can't argue against it being the more popular](https://trends.google.com/trends/explore?date=all&q=kebab-case,spinal-case,lisp-case,dash-case,caterpillar-case) 🤷🏻

PR #136, opened 2020-12-29
feat: generate tokens with theo

josokinas:
## Purpose

In order to (later) generate tokens not only for web, but also for iOS and Android, we need one source of aliases & tokens in a format that can be used to generate multi-platform variants.

## Approach

Using [Theo](https://github.com/salesforce-ux/theo) to generate (for now) web tokens.

I had an idea of running script each time either `aliases.json` or `tokens.json` gets modified, then commit the `tokens.scss` to Git, however due to issues I gave up, also not sure having `tokens.scss` in the repo would be much benefitial.

## Testing

Tested locally, pipeline should also pass, preview generated.

## Risks

N/A


josokinas:
The build is failing because workflow runs out of main branch, and this now has adjustment to generate the token. Once merged it will start passing.

scripts/formats/tokens.scss.hbs
```diff
@@ -0,0 +1,6 @@
+// generated by Theo (https://github.com/salesforce-ux/theo), edit `tokens.json` instead
+@mixin tokens() {
+{{#each props as |prop|}}
+  --ods-{{kebabcase prop.name}}: {{#eq prop.type "string"}}"{{/eq}}{{{prop.value}}}{{#eq prop.type "string"}}"{{/eq}};
```
rabelloo, line 4: We might be able to optimise the `json` file and the templates by following our own convention, e.g.

```json
{
  "type": "border-radius",
  "name": "small",
  "value": "{!border-radius-small}"
},
{
  "type": "color",
  "name": "neutral-600",
  "value": "{!color-neutral-600}"
},
```

```hbs
{{#each props as |prop|}}
  --ods-{{prop.type}}-{{kebabcase prop.name}}: {{prop.value}};
{{/each}}
```

If `type` is a predefined enum, we could use `category` instead.
Regardless, I don't think the `eq string` bit is needed at all.

scripts/generate.ts
```diff
@@ -0,0 +1,33 @@
...
+    file: type.src,
+  },
+  format: {
+    type: 'scss',
+  },
+})
+  .then((scss) => {
+    writeFileSync(type.dest, scss);
+  })
+  .catch((error) => console.error(error));
```
rabelloo, line 33: 💅🏻  nitpick

```suggestion
  .then((scss) => writeFileSync(type.dest, scss))
  .catch(console.error);
```

package.json
```diff
@@ -10,8 +10,10 @@
     "node": ">= 14"
   },
   "scripts": {
+    "prestart": "yarn generate",
     "start": "start-storybook -p 6006",
-    "prebuild": "shx rm -rf packages/*/dist",
+    "generate": "esbuild scripts/generate.ts --bundle --platform=node --target=es2019 --format=cjs --external:theo | node",
```
rabelloo, line 15: 1. Does `--external:theo` make it run faster?
I'm pretty sure I'd work the same without it.

2. `generate` is kinda generic.
Is there a name that would convey the meaning a bit better?
Perhaps `tokens` or `themes` or even `generate:tokens`.

rabelloo (approved):

scripts/formats/tokens.scss.hbs
josokinas: Good suggestion, however I feel that because "value" would still need to alias to full key it might be just more confusing than having little more of code:

```json
{
  "type": "size",
  "category": "border-radius",
  "name": "small",
  "value": "{!border-radius-small}"
}
```

vs

```json
{
  "type": "size",
  "category": "border-radius",
  "name": "border-radius-small",
  "value": "{!border-radius-small}"
}
```

package.json
josokinas: Yes, there will also have to be a theme(s) generation - changed to `generate:tokens`.

Regarding `--external`, it's needed to overcome problems such as: https://github.com/evanw/esbuild/issues/247, otherwise Theo does not run with TypeScript.

scripts/formats/tokens.scss.hbs
rabelloo: Don't you think it's less confusing since a `theo alias` will be the full variable name, while a `theo prop` will be more or less equal to a Figma token?

```json
{
  "aliases": {
    "color-neutral-500": "180, 186, 197"
  },
  "props": [
    {
      "type": "color",
      "name": "neutral-500",
      "value": "{!color-neutral-500}"
    },
    {
      "type": "color",
      "name": "content-main",
      "value": "{!color-neutral-500}"
    }
  ]
}
```

i.e. `{!color-neutral-500}` has to point to an alias, because the prop is `color/neutral-500`

package.json
rabelloo: Ah I see. Theo itself is not a proper esm and can't be bundled by esbuild properly. Cool.

scripts/formats/tokens.scss.hbs
josokinas: OK, refactored this bit - have a look.

I can't use `type` here as type has to be "size" for border radius, this for web does not mean much but is used e.g. for Android generation.

rabelloo: Looks good. I figured `type` was used some other way.
We might revert this decision in the future, but for now it looks way cleaner.

PR #175, opened 2021-01-12
docs: don't set "href" key when value is empty string

josokinas:
## Purpose

Button component uses `<a>` tag when "href" prop is present.

A story starts with it being `undefined`, however when entering any value into "href" input on the story and then removing it, "href" prop becomes an empty string instead of previous `undefined` value, and this does not allow to revert component to use `<button>` tag so "disabled" prop is not working anymore.

## Approach

Only set "href" prop to component when it has a value.

## Testing

On Storybook.

## Risks

N/A

### DIFF
--- a/packages/react/src/components/button/button.react.stories.tsx
+++ b/packages/react/src/components/button/button.react.stories.tsx
@@ -27,7 +27,17 @@ export default {
   },
 } as Meta<ButtonProps>;
 
-export const Playground: Story<ButtonProps> = (props) => <Button {...props} />;
+export const Playground: Story<ButtonProps> = ({
+  href,
+  ...restProps
+}: ButtonProps & Pick<JSX.IntrinsicElements['a'], 'href'>) => (
+  <Button
+    {...{
+      ...restProps,
+      ...(href && { href }), // don't set "href" key when value is empty string
+    }}
+  />
+);
 
 export const Kind = storyOf(Button, 'kind', ['action', 'destructive']);
 Kind.argTypes = omit<ButtonProps>('kind', 'children');


packages/react/src/components/button/button.react.stories.tsx
```diff
@@ -27,7 +27,17 @@ export default {
...
-export const Playground: Story<ButtonProps> = (props) => <Button {...props} />;
+export const Playground: Story<ButtonProps> = ({
+  href,
+  ...restProps
+}: ButtonProps & Pick<JSX.IntrinsicElements['a'], 'href'>) => (
+  <Button
+    {...{
+      ...restProps,
+      ...(href && { href }), // don't set "href" key when value is empty string
+    }}
```
rabelloo, line 38: I don't like using property spreads like this mainly for two reasons:
- TypeScript can't properly pick up the keys and allows things it shouldn't
- When one of the properties' types err, the message is really confusing and not helpful as it should be

e.g.
image

You can see it doesn't complain about the spread `href`, but does so correctly in the line below that.

I'd really prefer this: (a bit terser too)
```suggestion
    {...restProps}
    href={href || undefined} // don't set "href" when value is falsy
```

Oh and the comment change is a nitpick suggestion.

rabelloo (approved):

packages/react/src/components/button/button.react.stories.tsx
josokinas: Unfortunately this does not work...

Storybook in case of `undefined` argument does not forward it to props:

Screenshot 2021-01-12 at 12 53 33

But with this change `href` would always be there even `undefined`, and Button component explicitly checks for prop to be present (not for whether it is defined). This is because of how component overloads are designers.

rabelloo: 😭  🤷🏻

PR #176, opened 2021-01-12
docs: better describe textarea component "resize" variations

josokinas:
## Purpose

After [adding "resize" table options to Textarea component](https://github.com/onfido/castor/pull/157) it is now not aligned with other type description that allow only certain "resize" prop to be set:

Screenshot 2021-01-12 at 11 39 17

For example, on Button component we describe what "kind" prop can be set.

## Approach

Modify table options to describe available "resize" prop values:

Screenshot 2021-01-12 at 11 36 59

This also changes picker to "radio", same as used by default on Button component.

## Testing

On Storybook.

## Risks

Because this is not picked up from types by Storybook anymore (due to `withRef` wrapper usage), it needs to be manually aligned with any change made.

### DIFF
--- a/packages/react/src/components/textarea/textarea.react.stories.tsx
+++ b/packages/react/src/components/textarea/textarea.react.stories.tsx
@@ -2,6 +2,14 @@ import { Textarea, TextareaProps } from '@onfido/castor-react';
 import React from 'react';
 import { Meta, omit, Story, storyOf } from '../../../../../docs';
 
+const resizes: readonly TextareaProps['resize'][] = [
+  'vertical',
+  'horizontal',
+  'both',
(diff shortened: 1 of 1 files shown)


rabelloo (approved):
So annoying, is it because we wrap `Textarea` in `withRef` but not `Button`?

josokinas:
> ...is it because we wrap `Textarea` in `withRef` but not `Button`?

Indeed, I believe Storybook does not pick up types. There might be a better solution, did not investigate TBH...

rabelloo:
No worries, we can always come back to it.
Storybook's type inference for different file component/types is already not great 🤷🏻

Issue #151, opened 2021-01-07
Component name not shown on Storybook code snippet when withRef is used


josokinas:
When a component is using `withRef`, a name is no more visible on a code snippet of Storybook, e.g.:

<img width="1017" alt="Screenshot 2021-01-07 at 18 39 06" src="https://user-images.githubusercontent.com/20243687/103918685-02e33e00-5107-11eb-817b-2c111a8b27f3.png">

rabelloo:
Easy workaround is to add `displayName` on each component
```tsx
const Input = withRef(...);
Input.displayName = 'Input';
```

I don't think there's a way that `withRef` can do it without any explicit name giving, that'd be nice 😞 
But other ideas:

```tsx
const Input = withRef('Input', () => {});
const Input = withRef(function Input(){ });
```

PR #177, opened 2021-01-12
fix(react): withRef component displayNames

rabelloo:
## Purpose

Closes https://github.com/onfido/castor/issues/151

## Approach

Explicitly name functions used inside `withRef` so that the function's name is used as `displayName`.

## Testing

Go to Storybook, select their stories, click show code, confirm it shows the component name.

## Risks

There's no way to guarantee `withRef` will be used with named functions in the future.
Anonymous functions will continue to have the same issue.
There is a lint rule for `forwardRed`, but then the types get all messed up (which `withRef` fixes).

### DIFF
--- a/packages/react/src/utils/withRef.ts
+++ b/packages/react/src/utils/withRef.ts
@@ -6,6 +6,10 @@ import { FC, forwardRef } from 'react';
  * @param component Component to `forwardRef`.
  */
 /* eslint-disable @typescript-eslint/no-explicit-any */
-export const withRef = <C extends FC<any>>(component: C): C =>
+export const withRef = <C extends FC<any>>(component: C): C & Named =>
   forwardRef(component as any) as any;
 /* eslint-enable @typescript-eslint/no-explicit-any */
+
+interface Named {
+  displayName?: string;
+}
--- a/packages/react/src/components/input/input.react.tsx
+++ b/packages/react/src/components/input/input.react.tsx
@@ -2,19 +2,19 @@ import { c, classy, InputProps as BaseProps, m } from '@onfido/castor';
 import React from 'react';
 import { withRef } from '../../utils';
 
-export const Input = withRef(
-  (
-    { type = 'text', invalid, className, ...restProps }: InputProps,
-    ref: InputProps['ref']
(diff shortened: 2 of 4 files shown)


josokinas:
I think the solution here is fine, however how do you feel about setting `displayName` for when component uses `withRef`, e.g.:

```js
Textarea.displayName = 'Textarea';
```

josokinas (approved):

rabelloo:
> how do you feel about setting `displayName` for when component uses `withRef`

It'd require some more type juggling, plus the name isn't localised in a single line so it could make it easier to miss, but I don't mind that solution either.

I'll push another commit for comparison and you tell me what you think.

josokinas:
> I'll push another commit for comparison and you tell me what you think.

Ah, not needed - I'm fine with this! Thanks 😏

rabelloo:
> Ah, not needed - I'm fine with this! Thanks 😏

Too late, take a look now 🤣

packages/react/src/utils/withRef.ts
```diff
@@ -6,6 +6,10 @@ import { FC, forwardRef } from 'react';
  * @param component Component to `forwardRef`.
  */
 /* eslint-disable @typescript-eslint/no-explicit-any */
-export const withRef = <C extends FC<any>>(component: C): C =>
+export const withRef = <C extends FC<any>>(component: C): C & Named =>
   forwardRef(component as any) as any;
 /* eslint-enable @typescript-eslint/no-explicit-any */
+
+interface Named {
+  displayName?: string;
```
josokinas, line 14: Could this be not optional?

josokinas (approved):
Nice! I like with `displayName` more, but happy with both - you choose!

packages/react/src/utils/withRef.ts
rabelloo: It could, but it wouldn't be accurate, because the return from `withRef` has an undefined `displayName`.

rabelloo:
> Nice! I like with `displayName` more, but happy with both - you choose!

I'll leave as is, with this we can use either in the future.

PR #208, opened 2021-01-20
fix(core): input and textarea borders on focus

### LINKED ISSUE #207: Focus border color should override hover color
At the moment if you focus an textarea or input, and then hover it, the standard hover border color is applied. It should just stay as the focus color with no change on hover.

image

rabelloo:
## Purpose

Closes #207 

## Approach

Refactors their styles to be shared in a mixin `as-text-input`.
That mixin also fixes the border specificity issue,
by applying it between `:hover` and `.-invalid`,
which requires a `stylelint` line disable.

## Testing

Visually check the borders on focus.

## Risks

None.

### DIFF
--- a/packages/core/src/mixins/as-text-input.scss
+++ b/packages/core/src/mixins/as-text-input.scss
@@ -1,0 +1,45 @@
+@use '../helpers';
+@use './with-inner-focus' as mixins;
+
+@mixin as-text-input() {
+  @include helpers.font('300-regular');
+
+  $border-width: 1px;
+
+  background-color: helpers.color('background-input');
+  border: $border-width solid helpers.color('border-input');
+  border-radius: helpers.border-radius('medium');
+  box-sizing: border-box;
+  color: helpers.color('content-main');
+  padding: helpers.space(1.5) - $border-width helpers.space(2) - $border-width;
+  width: helpers.space(50);
+
+  &::placeholder {
+    color: helpers.color('content-placeholder');
+  }
+
+  &:hover {
+    border-color: helpers.color('border-input-hover');
+  }
+
+  // needs to be applied between [hover] and invalid states
+  // stylelint-disable-next-line order/order
+  @include mixins.with-inner-focus('action', $border-width);
+
+  &.-invalid {
+    @include mixins.with-inner-focus('negative', $border-width);
+
+    border-color: helpers.color('border-negative');
+    color: helpers.color('content-negative');
+  }
+
+  &:disabled {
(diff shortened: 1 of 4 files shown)


packages/core/src/mixins/as-text-input.scss
```diff
@@ -0,0 +1,44 @@
+@use '../helpers';
+@use './with-inner-focus' as mixins;
```
josokinas, line 2: [Minor] Why not still use mixins as before:

```suggestion
@use '../mixins';
```

packages/core/src/mixins/as-text-input.scss
```diff
@@ -0,0 +1,44 @@
...
+
+  &::placeholder {
+    color: helpers.color('content-placeholder');
+  }
+
+  &:hover {
+    border-color: helpers.color('border-input-hover');
+  }
+
+  // needs to be applied between :hover and .-invalid
```
josokinas, line 25: Just to keep consistency with Radio component:

```suggestion
  // needs to be applied between [hover] and invalid states
```

packages/core/src/mixins/as-text-input.scss
```diff
@@ -0,0 +1,44 @@
...
+  // stylelint-disable-next-line order/order
+  @include mixins.with-inner-focus('action', $border-width);
+
+  &.-invalid {
+    @include mixins.with-inner-focus('negative', $border-width);
+    border-color: helpers.color('border-negative');
+    color: helpers.color('content-negative');
```
josokinas, line 32: [Minor] Totally up to you here, but:

```suggestion
    @include mixins.with-inner-focus('negative', $border-width);
   
    border-color: helpers.color('border-negative');
    color: helpers.color('content-negative');
```

josokinas (approved):

packages/core/src/mixins/as-text-input.scss
rabelloo: Because it creates a bidirectional import path, or a loop.
`as-text-input` imports `mixins/index`, while `mixins/index` imports `as-text-input`.
Sass might be clever enough to make it work, but it's bad practice 🙊

rabelloo: Actually it doesn't work
image

josokinas: Yeah, it's OK - it was a minor comment, I do not require actions on these 😄

Thanks for checking though!

PR #209, opened 2021-01-20
fix: selectors applied conditionally

rabelloo:
## Purpose

Sass' `&` will output the selector twice,
which is why the linter disallows it,
and our conditional selectors should account for that.

image

## Approach

Apply with `@if` instead of function `if` which allows for better content control.
Create mixin for shareability.

## Testing

Check output CSS.

image

## Risks

Reorders stylelint rule, `@include` before `$variable`, which honestly is a better separation and allows local variables to be used in mixins.

### DIFF
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -11,15 +11,9 @@
     "scss/no-global-function-names": true,
     "order/order": [
       [
-        {
-          "type": "at-rule",
-          "name": "include"
-        },
-        {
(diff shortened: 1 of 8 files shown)


josokinas (approved):

josokinas:
P.S. Should this be `fix(core)`?

rabelloo:
Yea the PR title didn't update but I included `core` as scope in the commit

PR #215, opened 2021-01-22
chore: use any @onfido/castor-icons version

josokinas:
## Purpose

Dependabot does not currently upgrade majors of `@onfido/castor-icons`, however we actually can and need to have always the latest.

## Approach

Allow to use _any_ version of `@onfido/castor-icons`.

This also upgrades to the latest.

## Testing

After next major release of `@onfido/castor-icons` Dependabot should open a PR to upgrade.

## Risks

Not entirely sure if Dependabot won't change the `*` indicating _any_ version.

### DIFF
--- a/examples/custom-themes/package.json
+++ b/examples/custom-themes/package.json
@@ -8,7 +8,7 @@
     "start": "parcel src/index.html --open"
   },
   "dependencies": {
-    "@onfido/castor-icons": "^1.2.0"
+    "@onfido/castor-icons": "*"
   },
   "devDependencies": {
     "autoprefixer": "^9.8.6",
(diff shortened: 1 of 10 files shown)


rabelloo (approved):

rabelloo:
My only concern is people copying that from the examples and inadvertently introducing bugs in their app in the future when a breaking change is introduced.

josokinas:
> My only concern is people copying that from the examples and inadvertently introducing bugs in their app in the future when a breaking change is introduced.

I did not find other way for Dependabot, I think it's same as e.g. manually trying to run `yarn upgrade @onfido/castor-icons` - it won't upgrade.

But then, developer should know this basic thing about `package.json`...

rabelloo:
Yeah that's fine, I agree.

PR #224, opened 2021-01-25
refactor(core): use sass:list built-in module

josokinas:
## Purpose

Always use built-in Sass modules.

## Approach

Use `sass:list`  module instead of global function `index`.

## Testing

Workflow build should pass.

Unfortunetly, this is not enabled [on `no-global-function-names`stylelint rule](https://github.com/kristerkari/stylelint-scss/blob/deb3e6bbe2d4a526412eff9b60b9ca00192e05ff/src/rules/no-global-function-names/README.md), maybe because it's not part of official [full list of disallowed names](https://github.com/sass/sass/blob/master/accepted/module-system.md#built-in-modules-1).

## Risks

Nothing restricts from still using global `index`.

### DIFF
--- a/packages/core/src/helpers/border-radius/border-radius.scss
+++ b/packages/core/src/helpers/border-radius/border-radius.scss
@@ -1,7 +1,9 @@
+@use 'sass:list';
+
 $_allowed-sizes: ('small', 'medium', 'large', 'full');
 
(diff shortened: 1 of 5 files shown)


rabelloo (approved):
Looks good, too bad the linter doesn't catch it.

PR #225, opened 2021-01-25
feat(core)!: do not distribute individual component css files

josokinas:
## Purpose

For CSS build individual components will not be distributed, instead Sass is recommended to be used,
or dead code should be eliminated from "castor.css" file.

## Approach

Remove generation of individual component CSS files for distribution.

## Testing

On local build.

## Risks

This is a breaking change - indicating with a note.

### DIFF
--- a/packages/core/style/components/button.scss
+++ b/packages/core/style/components/button.scss
@@ -1,4 +1,0 @@
-// this file will generate ./dist/components/button.css
-@use '../../src/components';
-
-@include components.Button();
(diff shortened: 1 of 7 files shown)


rabelloo (approved):

rabelloo:
CSS dead code elimination can be performed by [PurgeCSS](https://github.com/FullHuman/purgecss).

We could document that along with `castor.css` usage at a later stage.

PR #231, opened 2021-01-26
fix(core): checkbox component "invalid" state styling

josokinas:
## Purpose

Styling for Checkbox "invalid" state is not right.

## Approach

Fix styling to apply correct "invalid" colors for [checked] and [indeterminate] indicators, also outline.

## Testing

On Storybook using "night" mode.

## Risks

N/A

### DIFF
--- a/packages/core/src/components/checkbox/checkbox.scss
+++ b/packages/core/src/components/checkbox/checkbox.scss
@@ -57,10 +57,19 @@
 
   // indicating invalid and not [disabled] input
   .ods-checkbox.-invalid:not(:disabled) {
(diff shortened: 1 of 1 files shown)


rabelloo (approved):
Code looks good but our tokens might be incorrect.

The preview looks different than the design.
Slightly different red colors, should be the same.

Screenshot 2021-01-26 at 17 19 23

josokinas:
> Code looks good but our tokens might be incorrect.

Checked with @Subcide, everything is fine. Design will be adjusted to reflect such state with having a border, where originally it did not.

PR #232, opened 2021-01-26
fix: use "gap" with grid

josokinas:
## Purpose

Using "gap" with Flexbox layout is not supported on iOS Safari, instead Grid layout is to be used.

## Approach

Use "grid" instead of "flex" (incl. "inline").

## Testing

On iOS Safari, and regular browser(s).

## Risks

N/A


rabelloo (approved):

rabelloo (changes requested):
Oh whoops, we missed one thing:
image

`grid-auto-flow: column;`

rabelloo (approved):

PR #234, opened 2021-01-27
build: bump "core" as peer dependency

josokinas:
## Purpose

When bumping a version, also bump "core" as peer dependency, in current case only "react" package.

## Approach

Use a [custom updater for `standard-version`](https://github.com/conventional-changelog/standard-version#can-i-use-standard-version-for-additional-metadata-files-languages-or-version-files) to bump `@onfido/castor` as a peer dependency of `@onfido/castor-react`.

Unfortunetly, as the script is invoked by `standard-version` itself, I cannot use TypeScript here.

## Testing

`yarn release --dry-run`

Screenshot 2021-01-27 at 15 28 04

## Risks

N/A

### DIFF
--- a/scripts/bump-core-peer-dependency.cjs
+++ b/scripts/bump-core-peer-dependency.cjs
@@ -1,0 +1,18 @@
+// can't be `bump-core-peer-dependency.ts` as it is invoked by
+// `standard-version` as a custom updater
+
+const detectIndent = require('detect-indent');
+const detectNewline = require('detect-newline');
+const stringifyPackage = require('stringify-package');
+
+module.exports.readVersion = function (contents) {
+  return JSON.parse(contents).peerDependencies['@onfido/castor'];
+};
+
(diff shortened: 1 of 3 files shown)


scripts/bump-core-peer-dependency.cjs
```diff
@@ -0,0 +1,18 @@
+// can't be `bump-core-peer-dependency.ts` as it is invoked by
```
rabelloo, line 1: It technically can be, if we compile it prior to `release` script's execution, for example:

`package.json`
```json
"scripts": {
  "prerelease": "esbuild bump-core-peer-dependency.ts --module=cjs > bump-core-peer-dependency.cjs",
```

`.gitignore`
```
bump-core-peer-dependency.cjs
```

But I don't mind keeping the `.cjs` file as it's super simple.

rabelloo (approved):

PR #236, opened 2021-01-27
fix(react): indicator data- and aria- properties apply to label

### LINKED ISSUE #230: Checkbox + Cypress fails to click `input`
### Issue
Cypress fails to click Checkbox's `<input />` because:

```html
<input name="noClearVideo-noFacePresent" type="checkbox" class="ods-checkbox">
is being covered by another element:
<span class="ods-input-indicator" aria-hidden="true"></span>
```

### Reproduce with:
```tsx
<Checkbox data-test="my-checkbox" />
```

```ts
cy.get('[data-test=my-checkbox]').click();
```

### Proposed solution
Either:
1. `data-` and `aria-` attributes need to be applied to `<label />` element instead of `<input />` which makes the click happen on a visible element
2. `<input />` needs to be visible on top of `<span />` which might include `z-index` manipulation and/or hacks

`1.` is probably more intuitive usage anyway, to unwrap attributes/props on the container instead of the input.
We could destructure `input`'s attributes out of `props` and pass the rest to `label`.

rabelloo:
## Purpose

closes #230

## Approach

Pick `data-` and `aria-` properties from `restProps` and apply to `IndicatorContainer` which renders to `label`.

This fixes the Cypress issue as any `data-test` attribute will be placed on the `label` which will then be clicked without issue.

## Testing

Apply `data-` attributes in a story and they should render in `label` instead of `input`.

## Risks

`partition` is implemented "imperatively", as in the data argument is the first.


packages/react/src/utils/fp/map/map.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Transforms each item in an `array` according to a function `mapFn`.
+ * @param mapFn A function that projects each member to its new form.
```
josokinas, line 3:
```suggestion
 * Transforms each item in an `array` according to a function `mapFn`.
 *
 * @param mapFn A function that projects each member to its new form.
```

packages/react/src/utils/fp/partition/partition.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * Partitions an `array` in two with elements that satisfy the `predicate`
+ * then those that don't, respectively.
```
josokinas, line 3:
```suggestion
 * Partitions an `array` in two with elements that satisfy the `predicate` then
 * those that don't, respectively.
```

packages/react/src/utils/fp/pipe/pipe.spec.ts
```diff
@@ -0,0 +1,16 @@
+import { describe, expect, it } from '@jest/globals';
+import { pipe } from './pipe';
+
+describe('pipe', () => {
+  it('should split the list in a tuple of elements that satisfy the predicate, then those that do not', () => {
```
josokinas, line 5: [Minor] Way too long description...

josokinas (changes requested):
I'm not sure about going towards functional programming way - why did you choose to do such?

Previously when we tried to introduce such in other projects it was confusing to developers. It is not an easy way of coding, and I would prefer to avoid if possible.

Also, I feel like creating all tooling is also not right, if you do really want to go fp, then probably using [Ramda](https://ramdajs.com/) is better.

rabelloo:
> I'm not sure about going towards functional programming way - why did you choose to do such?
> 
> Previously when we tried to introduce such in other projects it was confusing to developers. It is not an easy way of coding, and I would prefer to avoid if possible.
> 
> Also, I feel like creating all tooling is also not right, if you do really want to go fp, then probably using [Ramda](https://ramdajs.com/) is better.

It read really poorly without a nice `pipe`.

Although Ramda is a nice library, I'd still prefer we remain dependency free, and these functions are simple to recreate individually as we need them.

I'll refactor to no FP then, intermediate variables and all, for a re-review.

packages/react/src/internal/indicator-container/splitContainerProps.ts
```diff
@@ -0,0 +1,18 @@
+import { map, partition, pipe } from '../../utils';
+
+/**
+ * Splits `Indicator` properties into `label` and `input` props.
+ */
+export const splitContainerProps = <T extends Input | Label>(props: T) =>
```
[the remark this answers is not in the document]
rabelloo: Right, I don't mind making it more generic now, do you want me to do that?
I'd still wrap this more specific one though to make sure they're shared between Checkbox and Radio.

Usage would look something like this:
```ts
const containerProps = /^data-|^aria-/;
const splitContainerProps = splitProps<Partial<Label>>(([key]) => containerProps.test(key));

const [myProps, otherProps] = splitContainerProps(restProps);
// myProps: Partial<Label>
// otherProps: Partial<typeof restProps>
```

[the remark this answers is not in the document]
josokinas: It's up to you, no need to take action now.

packages/react/src/utils/partition/partition.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * Partitions an `array` in two with elements that satisfy the `predicate` then
+ * those that don't, respectively.
+ */
```
josokinas, line 4: I guess good thing to add `@param` docs here too?

packages/react/src/internal/indicator-container/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './indicator-container';
```
josokinas, line 1: Personally instead of adding an index, I would include the utility in the `indicator-container.tsx`, and export it from there.

This is because the only way you'd want to use this is when using an internal component itself.

At the end, they both would still be exported from `utils`.

It's up to you to decide which way to leave here.

josokinas (approved):

josokinas:
> I'll refactor to no FP then, intermediate variables and all, for a re-review.

To me this actually is easier to understand, but feel free to ask for someone else's opinion too.

packages/react/src/internal/indicator-container/index.ts
rabelloo: In general I like single export files, makes testing easier and better scoped, and maintainability simpler.
In special separating `.tsx` from `.ts` files is extra helpful because the parser doesn't get confused by `<T>` types.

Lastly, I just realised `splitContainerProps.ts` is missing unit tests, our `jest` config ignores everything inside `internal`.
I'll fix that, that'll probably make my arguments clearer.

Don't mind later removing `internal/indicator-container/index.ts` in favour of `internal/index.ts` though.

josokinas: Yes, indeed configuration now ignores all `components/**.*`.

What I meant is:

```
internal/indicator-container/indicator-container.tsx
internal/indicator-container/splitContainerProps.ts
```

Then `indicator-container.tsx`:

```ts
import { splitContainerProps } from './splitContainerProps.ts'

const IndicatorContainer = ...

export {
  IndicatorContainer,
  splitContainerProps,
};
```

rabelloo:
> To me this actually is easier to understand, but feel free to ask for someone else's opinion too.

No that's fine, unfortunately for me functional hasn't picked up as much as imperative yet 😢 
It's not like JavaScript has good language constructs for FP either, we have to work around it.

packages/react/src/internal/indicator-container/splitContainerProps.ts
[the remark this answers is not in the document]
rabelloo: Generalised `partitionObject`.

packages/react/src/internal/indicator-container/index.ts
rabelloo: I see, that's a bit weird but I'll give it a try.

packages/react/src/internal/indicator-container/splitContainerProps.ts
```diff
@@ -0,0 +1,12 @@
...
+/**
+ * Splits `Indicator` properties into `label` and `input` props.
+ */
+export const splitContainerProps = (props: Input | Label) =>
+  partitionObject(props, ([key]) => containerProps.test(key));
+
+const containerProps = /^aria-|^data-/;
+
+type Input = JSX.IntrinsicElements['input'];
+type Label = JSX.IntrinsicElements['label'];
```
josokinas, line 12: In other places we have different naming for element props, what do you think about:

```suggestion
export const splitContainerProps = (props: InputElementProps | LabelElementProps) =>
  partitionObject(props, ([key]) => containerProps.test(key));

const containerProps = /^aria-|^data-/;

type InputElementProps = JSX.IntrinsicElements['input'];
type LabelElementProps = JSX.IntrinsicElements['label'];
```

rabelloo: Yup I'll keep the consistency, even though the verbosity drives me nuts 😅

packages/react/src/utils/partitionObject/partitionObject.spec.ts
```diff
@@ -0,0 +1,13 @@
+import { describe, expect, it } from '@jest/globals';
+import { partitionObject } from './partitionObject';
+
+describe('partitionObject', () => {
+  it('should split an object like [{ "does" }, { "does not" }] satisfy the predicate', () => {
+    const obj = { foo: true, bar: false };
+
+    const [truey, falsey] = partitionObject(obj, Boolean);
```
josokinas, line 8: Very interesting naming, what about `truthy` and `falsy`?

rabelloo: 🙈  renamed

yarn.lock
```diff
@@ -7192,9 +7192,9 @@ falafel@^2.1.0:
     isarray "^2.0.1"
     object-keys "^1.0.6"
 
-"falsey@^0.3.2":
+"falsy@^0.3.2":
```
josokinas, line 7195: Probably a mistake?

packages/react/src/utils/partition/partition.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * Partitions an `array` in two with elements that satisfy the `predicate` then
+ * those that don't, respectively.
+ *
+ * @param array List to be partitioned.
+ * @param predicate Function that determines in where to put each element.
+ */
+export const partition = <T>(
```
josokinas, line 8: Do you see this being used elsewhere rather than for `partitionObject`?

rabelloo: Not really, maybe for something like separating tokens perhaps

```ts
partition([...colors, ...borderRadii], (token) => token.startsWith('ods-color-'));
```

yarn.lock
rabelloo: 🙈  yeap, thanks

packages/react/src/utils/partition/partition.ts
josokinas: In that case maybe it could go inside `partitionObject.js` file?

It's a little confusing, to me a utility is something that can be (and will be) used in multiple places.

Separating props is a such utility. But maybe if these two are merged into one file, and final utility is called `partitionProps`?

What do you think?

rabelloo: It's a bit preparing for the future I suppose.
The only thing I don't like about `partitionProps` is that it can work on any object, not just props.
I don't mind removing partition for now though, we can always reintroduce later when needed.

PR #239, opened 2021-01-28
docs(react): inform to include "icons" sprite

josokinas:
## Purpose

Make it clear to include `Icons` SVG sprite when using React `Icon` component.

## Approach

Add instructions both on README ("react" package) and Storybook (React section).

## Testing

On GitLab itself, then on Storybook preview.

## Risks

N/A

### DIFF
--- a/docs/_react/intro.story.tsx
+++ b/docs/_react/intro.story.tsx
@@ -32,5 +32,24 @@ export const Intro = () => (
       format
       dark
     />
+    <p>
+      Then (only once) inline the <code>Icons</code> SVG sprite in your app:
+    </p>
+    <Source
+      language="jsx"
+      code={`
+        import { Icons } from '@onfido/castor-icons';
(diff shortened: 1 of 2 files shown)


docs/_react/intro.story.tsx
```diff
@@ -32,5 +32,24 @@ export const Intro = () => (
       format
       dark
     />
+    <p>
+      Then (only once) inline the <code>Icons</code> SVG sprite in your app:
```
rabelloo, line 36: We seem to assume they are going to use icons.
Should we perhaps include language for "if you are going to use <Icon />" for this paragraph and for installing `@onfido/castor-icons`?

rabelloo (approved):

docs/_react/intro.story.tsx
josokinas: Good point.

I will do this as separate PR, as it also includes docs for "core" part.

PR #241, opened 2021-01-28
docs(react): inform that old browsers are not supported

josokinas:
## Purpose

Clearly inform that old browsers are not supported.

## Approach

When mentioning IE11, also inform that Castor does not support old browsers.

## Testing

On GitHub itlsef.

## Risks

We do not actually have a list of supported browsers.

### DIFF
--- a/packages/react/README.md
+++ b/packages/react/README.md
@@ -40,7 +40,7 @@ Please note that Castor is exported as an [ECMAScript module](https://developer.
 
 You may wish to configure your bundler to transpile to a different module syntax, and/or to target lower a ECMAScript version.
 
-For example, you might choose UMD module syntax targeting ES5 if your app needs to support IE11.
+For example, you might choose UMD module syntax targeting ES5 if your app needs to support IE11 (please note that Castor is not tested in Internet Explorer).
 
 ### Make custom styled components
 


packages/react/README.md
```diff
@@ -40,7 +40,7 @@ Please note that Castor is exported as an [ECMAScript module](https://developer.
 
 You may wish to configure your bundler to transpile to a different module syntax, and/or to target lower a ECMAScript version.
 
-For example, you might choose UMD module syntax targeting ES5 if your app needs to support IE11.
+For example, you might choose UMD module syntax targeting ES5 if your app needs to support IE11 (please note that we do not support old browsers - use at your own risk).
```
rabelloo, line 43: I don't like the `use at your own risk` part, feels like saying there is in fact an inherent risk.

Also, `old` is subjective, what browsers do we support?
Should we have a dedicated section to list them?

josokinas: So I mentioned this in the description, that we actually _do not_ have a list of supported browsers.

Something to think about? Probably we should.

What do you think about just "please note that we do not test Castor against IE11 or previous versions".

rabelloo: Yea we probably should, and create a `.browserslist` file too.

That's fine, or `please note that Castor is not tested in Internet Explorer`.

rabelloo (approved):

packages/react/README.md
josokinas: https://github.com/onfido/castor/issues/243

PR #242, opened 2021-01-29
docs: improve installation instructions of icons on storybook

josokinas:
## Purpose

Storybook installation instructions are not clear enough on optional usage of  `@onfido/castor-icons`.

## Approach

Instruct on how to additionally install `@onfido/castor-icons`.

Also, leave only instructions using npm, as Yarn is something that is individual per app and setup should be obvious based on their API.

## Testing

On Storybook.

## Risks

N/A

### DIFF
--- a/docs/_core/intro.story.tsx
+++ b/docs/_core/intro.story.tsx
@@ -10,23 +10,23 @@ export const Intro = () => (
 
     <h2>Install</h2>
     <p>
-      You can install with <code>npm</code>:
+      You can install with <code>npm</code> CLI tool:
     </p>
     <Source
       language="bash"
       code={`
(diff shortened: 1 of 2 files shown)


docs/_core/intro.story.tsx
```diff
@@ -10,23 +10,23 @@ export const Intro = () => (
 
     <h2>Install</h2>
     <p>
-      You can install with <code>npm</code>:
+      You can install from <code>npm</code> registry:
```
rabelloo, line 13: There are many mirror registries like `yarn`'s.
I think it's better to say "install **with** the `npm` cli tool", the registry doesn't matter too much.

rabelloo (approved):

PR #249, opened 2021-02-01
feat: html stories support

### LINKED ISSUE #178: HTML + CSS stories
Implement HTML + CSS (vanilla) stories in `@onfido/castor` so that we can test styles in isolation without React/JS/state in the way.

rabelloo:
## Purpose

Closes #178

## Approach

Make use of docs `prepareForInline` to intercept HTML (`string` or `Element`) stories and render as React elements.

Also has `matrix` bundled in which is the replacement for our custom `storiesOf`, but could be split as a separate PR later.

## Testing

Open preview deploy to see new stories `core/Button` and `core/Icon`.

## Risks

Storybook canvas view is not supported.


packages/core/src/components/button/button.stories.ts
```diff
@@ -0,0 +1,42 @@
...
+    disabled: false,
+    kind: 'action',
+    variant: 'primary',
+  },
+} as Meta<HtmlButtonProps>;
+
+export const Playground: Story<HtmlButtonProps> = (props) => Button(props);
+
+export const AsAnchor: Story<HtmlButtonProps> = (props) =>
+  Button({ ...props, href: 'javascript:0' });
```
josokinas, line 19: This does not work, probably better to set as:

```suggestion
  Button({ ...props, href: '#' });
```

But then (in general for a button) disable `onClick`?

josokinas (changes requested):
Good job!

FYI "Canvas" tab does not work, you're aware of this?

What would be your plan, once this is finised, merge and then work on other components, or continue with same PR for those also?

packages/core/src/components/button/button.stories.ts
rabelloo: What doesn't work? It renders as `<a>` correctly and you can click it (has `:hover`, `:active` and other pseudos).

The annoying part about disabling `onClick` is that it shows on the code snippet, which I don't particularly like.

packages/core/src/components/button/button.story.tsx
```diff
@@ -0,0 +1,15 @@
...
+import { c, classy, m } from '../../utils';
+import { ButtonProps } from './button';
+
+export interface HtmlButtonProps extends ButtonProps {
+  children: string;
+  href?: string;
+}
+
+export const Button = ({ href, kind, variant, ...props }: HtmlButtonProps) =>
+  html(href ? 'a' : 'button', {
```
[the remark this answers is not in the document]
rabelloo: Hmm probably yea. I'll fix that.

packages/core/src/components/button/button.ts
```diff
@@ -1,4 +1,5 @@
 export interface ButtonProps {
+  disabled?: boolean;
   kind?: 'action' | 'destructive';
   variant?: 'primary' | 'secondary' | 'tertiary';
```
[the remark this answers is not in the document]
rabelloo: I'd prefer alphabetical everywhere, as other sorts are subjective and can't be linted.

However I don't mind having a subjective (by importance?) sort, we just have to manually check it every time 🤷🏻 

I'll apply the suggestion for now, we can decide on an interface member alphabetical sort lint rule later.

rabelloo:
> Good job!

Thanks :)

> FYI "Canvas" tab does not work, you're aware of this?

Yea listed under Risks.
There's probably a way to intercept all renderings, not just docs' with `prepareForInline`, but it's not well documented and I didn't look into storybook's internal code too much.

> What would be your plan, once this is finised, merge and then work on other components, or continue with same PR for those also?

This is more of a Proof of Concept, so I would work on other stories later.
I'd also rework the current stories so they match React's, I just wanted to see wether the prop matrix for future visual regression tests would work too.

packages/core/src/components/button/button.stories.ts
josokinas: Yes, but when I click it I end up on an empty page with `0` rendered 😓 (on Firefox).

rabelloo: Oh Firefox, you silly goose. `javascript: void 0;` should do the trick then.

```suggestion
  Button({ ...props, href: 'javascript:void 0' });
```

josokinas: Did it work for you? Does not for me...

rabelloo: For some reason the CI was not working yesterday, GitHub actions were down or something.
Try now.

josokinas: Yep, now it's all fine.

Probably would also be good to adjust React story too...

rabelloo: Done.

.storybook/custom/prepareForInline.tsx
```diff
@@ -0,0 +1,31 @@
...
+  if (content instanceof Element) return decorate(<DomNode>{content}</DomNode>);
+
+  return story;
+};
+
+const DomNode = ({ children }: { children: Element }) => (
+  <Html>{children.innerHTML}</Html>
+);
+
+const Html = ({ children }: { children: string }) => (
```
josokinas, line 25: Probably minor, however this seems like you're rendering `<html>` but of course you _are not_.

docs/decorators/container.scss
```diff
@@ -0,0 +1,29 @@
...
+  display: inline-grid;
+  gap: $space;
+}
+
+.flex {
+  align-items: center;
+  display: flex;
+  flex-flow: row wrap;
+
+  // --- this is the same as gap, except Safari supports it
```
josokinas, line 15: Curious, were you not able to use grid here?

josokinas:
Something's not right now with latest updates:

Screenshot 2021-02-12 at 10 29 28

packages/react/src/components/icon/icon.stories.tsx
```diff
@@ -1,35 +1,30 @@
...
-export const Playground: Story<IconProps> = (props: IconProps) => (
-  <Icon {...props} />
-);
+export const Playground: Story<IconProps> = (props) => <Icon {...props} />;
 
-export const Name = storyOf(Icon, 'name', iconNames, {
-  display: 'grid',
-  labelMode: 'after',
-});
+export const Name = reactMatrix(Icon, { name: iconNames }, (props) => (
```
josokinas, line 21: I have noticed that on core stories of a list of all icons, a code example shows icon code + icon name:

Screenshot 2021-02-12 at 10 39 37

But for React it only shows icon code:

Screenshot 2021-02-12 at 10 38 05

Is this what you were expecting?

packages/react/src/components/icon/icon.stories.tsx
```diff
@@ -1,35 +1,30 @@
...
+import { colors, Meta, omit, reactMatrix, Story } from '../../../../../docs';
 
 const [firstIconName] = iconNames;
 
 export default {
   title: 'React/Icon',
   component: Icon,
   argTypes: {
-    ...omit<IconProps>('className', 'style'),
-    name: { control: { type: 'select', options: iconNames } },
```
josokinas, line 19: Curious, so "name" now gets rendered based on type right? What about "color", why that does not work then?

packages/core/src/components/icon/icon.story.tsx
```diff
@@ -0,0 +1,22 @@
+import { IconProps } from '@onfido/castor';
+import { html } from '../../../../../docs';
+import { c, classy } from '../../utils';
+
+/**
+ * `.ods-icon` requires SVG sprite to be included in your app. If you wish to
+ * use individual SVGs, import them directly from `@onfido/castor-icons/svg`
+ * instead.
+ *
+ * https://github.com/onfido/castor-icons
```
josokinas, line 10: Tha'ts actually not accurate.

If individual SVGs are used, you won't even be able to use only `.ods-icon`, instead you will have to use mask etc.

I would discourage such as it's an advanced usage.

```suggestion
 * `.ods-icon` requires SVG sprite to be included in your app.
 *
 * https://github.com/onfido/castor-icons
```

packages/core/src/components/icon/icon.story.tsx
```diff
@@ -0,0 +1,22 @@
...
+ */
+export const Icon = ({ name }: IconProps) =>
+  html('svg', {
+    class: classy(c('icon')),
+    fill: 'currentColor',
```
josokinas, line 15: I believe you're missing "color" change here:

```suggestion
export const Icon = ({ name, color: token }: IconProps) =>
  html('svg', {
    class: classy(c('icon')),
    fill: token ? color(token) : 'currentColor',
```

packages/react/src/components/button/button.stories.tsx
```diff
@@ -92,3 +78,16 @@ WithIcon.argTypes = {
 WithIcon.args = {
   iconName: firstIconName,
 };
+
+export const AllCombinations = reactMatrix(
```
josokinas, line 82: Similar to Icon component, probably minor but also noticed that core stories render exact button children text what is being placed, however React only renders "Button":

Screenshot 2021-02-12 at 11 09 42

packages/react/src/components/button/button.stories.tsx
```diff
@@ -1,89 +1,75 @@
...
+import React from 'react';
+import { Meta, omit, reactMatrix, Story } from '../../../../../docs';
 
 export default {
   title: 'React/Button',
   component: Button,
   argTypes: {
-    ...omit<ButtonProps>('className', 'onClick', 'style'),
-    children: { control: 'text' },
+    children: {},
```
josokinas, line 10: Do this need to be added, looks like it's just an empty object?

packages/core/src/components/button/button.stories.ts
```diff
@@ -0,0 +1,59 @@
...
+export const Variant = htmlMatrix(Button, { variant });
+Variant.argTypes = omit<ButtonProps>('variant', 'children');
+
+export const Disabled = htmlMatrix(Button, { disabled });
+Disabled.argTypes = omit<ButtonProps>('disabled', 'children');
+
+export const AsAnchor: Story<ButtonProps> = (props) =>
+  Button({ ...props, href: 'javascript:void 0' });
+
+const icon = Icon({ name: 'alarm' });
```
josokinas, line 43: To make sure first icon is selected from available, also would always match React story:

```suggestion
const [firstIconName] = iconNames;
const icon = Icon({ name: firstIconName });
```

josokinas (changes requested):
I like this, some comments added though.

Probably downside here is that it complicates component development, that is more stories to write. But at the end we will get a good testing infrastructure (when such added).

.storybook/custom/prepareForInline.tsx
rabelloo: Ah, I see that now. Do you have a name suggestion? `<HtmlString>`? `<InnerHtml>`? `<RenderHtml>`?

packages/react/src/components/icon/icon.stories.tsx
rabelloo: Yea I don't transform React source code.
I could if you think that's valuable, we could remove `key` too for instance.

packages/react/src/components/icon/icon.stories.tsx
rabelloo: Correct, all the possible name values are picked up by storybook/react-docgen.
I don't know why it doesn't work for color, probably because the types are imported from another library?

packages/react/src/components/button/button.stories.tsx
rabelloo: All it does is reorder it in the table. We could remove it if you don't mind it being after other props.

rabelloo:
> I like this, some comments added though.
> 
> Probably downside here is that it complicates component development, that is more stories to write. But at the end we will get a good testing infrastructure (when such added).

Thanks for the patience and thoroughness.

We don't necessarily need to write HTML/core stories yet, but ultimately it will benefit our consumers so at some point we should.

The benefits for testing can be reaped from the React stories alone.

packages/react/src/components/button/button.stories.tsx
rabelloo: That is interesting, I would call this a bug, I'll take a look.

packages/core/src/components/icon/icon.story.tsx
rabelloo: Nice catch!

.storybook/custom/prepareForInline.tsx
josokinas: Could it be... `Container`? Or `InnerHtmlContainer`?

packages/react/src/components/button/button.stories.tsx
josokinas: I believe originally I did put this also for same reason. Well, I think also because it wasn't as "text" by default too.

I think I don't mind it being _after_, still in other stories we already have a mess of argument ordering so TBH no point of fighting...

packages/react/src/components/icon/icon.stories.tsx
josokinas: OK, cool, no worries then - it can be explicitly set here.

packages/react/src/components/icon/icon.stories.tsx
josokinas: I actually think both "key" here and also text on core stories are not needed (especially those - why these would be handy?) 🤔

Also, these to me are minors. If too difficult to remove - then it is fine.

rabelloo: I don't think it's too difficult, it would just mean running prettier on React code too.
I'll see what I can do.

.storybook/custom/prepareForInline.tsx
rabelloo: Renamed to `HtmlContainer`, as `inner` is not necessarily accurate.

packages/react/src/components/button/button.stories.tsx
rabelloo: Removed hard-coded ordering, we can look into general story arguments ordering later.

rabelloo:
> Something's not right now with latest updates

Fixed that bug, requires some explicit `parameters: { display: 'flex' }` for stories now, but I think that's the most obvious way to use it.

I could automatically apply that display to a Story created with `matrix` but it'd be kinda magical, and coupled to `withContainer`.

packages/react/src/components/button/button.stories.tsx
rabelloo: Weird and interesting one, fixed by cutting the `reactMatrix` features a little bit.
Namely the third argument is no longer a Function Component but simply a function that interpolates props.

packages/react/src/components/icon/icon.stories.tsx
rabelloo: Removing `key` from React source now, and fixed the bug alongside the `children` for Button below.
Now it consistently shows the actual source, however it does look weird:
image

I could support per-story source transform, but I guess it's ok to have actual source code show up?
Less magic?

josokinas: Probably yes, just that `{' '}` looks odd.

rabelloo: Now also replacing `{' '}` instances.
The line break should do the same thing.

docs/helpers/html.ts
```diff
@@ -0,0 +1,26 @@
+/**
+ * Renders as HTML inside a React `div`.
+
+ * @param name Tag name, to use as root element.
+ * @param param1 Properties and attributes.
```
josokinas, line 5: What's `param1`?

josokinas (approved):

docs/helpers/html.ts
rabelloo: The unnamed second argument, since I'm destructuring it inline.
It goes `function (param0, param1, param2, ...) {}`

The only other way to document that argument is to name it, then destructure in the function's body.

rabelloo:
Oh noes, I force pushed an old version 🤦🏻

rabelloo:
Phew not too bad, only missed one commit. Saved!

Issue #4, opened 2020-11-30
Field Label


josokinas:
Implement the Field Label component.

PR #267, opened 2021-02-04
feat: add field-label component

josokinas:
## Purpose

Add FieldLabel component.

## Approach

Implements FieldLabel component.

This closes issue #4.

## Testing

On Storybook preview.

## Risks

N/A

### DIFF
--- a/docs/spacer/react.tsx
+++ b/docs/spacer/react.tsx
@@ -17,10 +17,10 @@ interface Node {
 
 type NodeChildren = string | Node | Node[];
 
-function hasSpacer(node?: Node): boolean {
-  if (node?.type?.displayName === (Spacer as FC).displayName) return true;
+function hasSpacer(node: Node): boolean {
+  if (node.type?.displayName === (Spacer as FC).displayName) return true;
 
-  const children = node?.props?.children;
+  const children = node.props?.children;
 
   if (!children || typeof children === 'string') return false;
 
--- a/docs/spacer/react.tsx
+++ b/docs/spacer/react.tsx
@@ -17,10 +17,10 @@ interface Node {
 
 type NodeChildren = string | Node | Node[];
 
-function hasSpacer(node: Node): boolean {
-  if (node.type?.displayName === (Spacer as FC).displayName) return true;
+function hasSpacer(node?: Node): boolean {
+  if (node?.type?.displayName === (Spacer as FC).displayName) return true;
 
-  const children = node.props?.children;
+  const children = node?.props?.children;
 
   if (!children || typeof children === 'string') return false;
 
--- a/packages/core/src/components/field-label/field-label.scss
+++ b/packages/core/src/components/field-label/field-label.scss
@@ -5,8 +5,15 @@
     @include helpers.font('300-regular');
 
     color: helpers.color('content-main');
-    display: grid;
-    gap: helpers.space(0.5);
+    display: flex;
+    flex-flow: column wrap;
     padding-top: helpers.space(0.5);
+
+    + input,
+    + textarea,
+    > input,
+    > textarea {
+      margin-top: helpers.space(0.5);
+    }
   }
 }
--- a/packages/core/src/components/field-label/field-label.scss
+++ b/packages/core/src/components/field-label/field-label.scss
@@ -5,8 +5,8 @@
     @include helpers.font('300-regular');
 
     color: helpers.color('content-main');
-    display: flex;
-    flex-flow: column wrap;
-    padding: helpers.space(0.5) 0;
+    display: grid;
+    gap: helpers.space(0.5);
+    padding-top: helpers.space(0.5);
   }
 }
--- a/packages/core/src/components/field-label/field-label.scss
+++ b/packages/core/src/components/field-label/field-label.scss
@@ -9,10 +9,10 @@
     flex-flow: column wrap;
     padding-top: helpers.space(0.5);
 
-    + input,
-    + textarea,
-    > input,
-    > textarea {
+    + .ods-input,
+    + .ods-textarea,
+    > .ods-input,
+    > .ods-textarea {
       margin-top: helpers.space(0.5);
     }
   }
--- a/packages/core/src/components/field-label/field-label.scss
+++ b/packages/core/src/components/field-label/field-label.scss
@@ -1,0 +1,15 @@
+@use '../../helpers';
+
+@mixin FieldLabel() {
+  $padding: helpers.space(0.5);
+
+  .ods-field-label {
+    @include helpers.font('300-regular');
+
+    color: helpers.color('content-main');
+    display: flex;
+    flex-flow: column wrap;
+    padding-bottom: $padding;
+    padding-top: $padding;
+  }
+}
--- a/packages/core/src/components/field-label/field-label.scss
+++ b/packages/core/src/components/field-label/field-label.scss
@@ -1,15 +1,12 @@
 @use '../../helpers';
 
 @mixin FieldLabel() {
-  $padding: helpers.space(0.5);
-
   .ods-field-label {
     @include helpers.font('300-regular');
 
     color: helpers.color('content-main');
     display: flex;
     flex-flow: column wrap;
-    padding-bottom: $padding;
-    padding-top: $padding;
+    padding: helpers.space(0.5) 0;
   }
 }
--- a/packages/react/src/components/field-label/field-label.stories.tsx
+++ b/packages/react/src/components/field-label/field-label.stories.tsx
@@ -1,0 +1,40 @@
+import { FieldLabel, FieldLabelProps, HelperText } from '@onfido/castor-react';
+import React from 'react';
+import { Meta, omit, Story } from '../../../../../docs';
+
+export default {
+  title: 'React/FieldLabel',
+  component: FieldLabel,
+  argTypes: {
+    ...omit<FieldLabelProps>('className', 'style'),
+    children: { control: 'text' },
+  },
+  args: {
+    children: 'Label',
+  },
+} as Meta<FieldLabelProps>;
+
+export const Playground: Story<FieldLabelProps> = (props: FieldLabelProps) => (
+  <FieldLabel {...props} />
+);
+
+interface FieldLabelWithHelperTextProps extends FieldLabelProps {
+  label: string;
+  helperText: string;
+}
+
+export const WithHelperText = ({
+  label,
+  helperText,
+  ...restFieldLabelProps
+}: FieldLabelWithHelperTextProps) => (
+  <FieldLabel {...restFieldLabelProps}>
+    {label}
+    <HelperText>{helperText}</HelperText>
+  </FieldLabel>
+);
+WithHelperText.argTypes = omit<FieldLabelWithHelperTextProps>('children');
+WithHelperText.args = {
+  label: 'Label',
+  helperText: 'Helper text',
+};
--- a/packages/react/src/components/field-label/field-label.stories.tsx
+++ b/packages/react/src/components/field-label/field-label.stories.tsx
@@ -1,4 +1,10 @@
-import { FieldLabel, FieldLabelProps, HelperText } from '@onfido/castor-react';
+import {
+  FieldLabel,
+  FieldLabelProps,
+  HelperText,
+  Input,
+  Textarea,
+} from '@onfido/castor-react';
 import React from 'react';
 import { Meta, omit, Story } from '../../../../../docs';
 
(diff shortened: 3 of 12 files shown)


docs/spacer/react.tsx
```diff
@@ -17,10 +17,10 @@ interface Node {
 
 type NodeChildren = string | Node | Node[];
 
-function hasSpacer(node: Node): boolean {
-  if (node.type?.displayName === (Spacer as FC).displayName) return true;
+function hasSpacer(node?: Node): boolean {
```
josokinas, line 20: The change here is to allow conditional rendering in stories, like `{someCondition && <some-element />}`.

rabelloo: We actually never use `<Spacer>` inside a story, we can later simplify this decorator.

packages/core/src/components/field-label/field-label.scss
```diff
@@ -0,0 +1,15 @@
...
+  $padding: helpers.space(0.5);
+
+  .ods-field-label {
+    @include helpers.font('300-regular');
+
+    color: helpers.color('content-main');
+    display: flex;
+    flex-flow: column wrap;
+    padding-bottom: $padding;
+    padding-top: $padding;
```
rabelloo, line 13: 💅🏻 
```suggestion
    padding: helpers.space(0.5) 0;
```

packages/react/src/components/input/input.stories.tsx
```diff
@@ -37,3 +42,29 @@ export const Disabled = storyOf(Input, 'disabled', [true, false], {
...
+  label,
+  helperText,
+  ...restInputProps
+}: InputWithFieldLabelProps) => (
+  <>
+    <FieldLabel htmlFor={id}>
+      {label}
+      {helperText && <HelperText>{helperText}</HelperText>}
+    </FieldLabel>
+    <Input {...restInputProps} id={id} />
```
rabelloo, line 63: My ideal API would be

```tsx
<Field>
  My label
  <Input />
</Field>
```

We could then either use the provided `id` and `htmlFor` or auto generate one by default.

That complicates the styling though, we could use `gap` if we require the `<Input>` to be inside the `<Field>` which honestly makes more sense in my opinion.

docs/spacer/react.tsx
rabelloo: We actually never use <Spacer> inside a story, we can later simplify this decorator.

packages/core/src/components/field-label/field-label.scss
```diff
@@ -0,0 +1,19 @@
...
+
+    color: helpers.color('content-main');
+    display: flex;
+    flex-flow: column wrap;
+    padding-top: helpers.space(0.5);
+
+    + input,
+    + textarea,
+    > input,
+    > textarea {
```
rabelloo, line 15: Shouldn't these be our classes?

```suggestion
    + .ods-input,
    + .ods-textarea,
    > .ods-input,
    > .ods-textarea {
```

packages/core/src/components/field-label/field-label.scss
```diff
@@ -0,0 +1,19 @@
+@use '../../helpers';
+
+@mixin FieldLabel() {
+  .ods-field-label {
+    @include helpers.font('300-regular');
+
+    color: helpers.color('content-main');
+    display: flex;
+    flex-flow: column wrap;
+    padding-top: helpers.space(0.5);
```
rabelloo, line 10: The example below will incorrectly have 4px padding top:

```tsx
<label class="ods-field-label">
  <input class="ods-input" />
</label>
```

Do we just not support that usage? Or don't care about that "error"?

Could be fixed with `gap` but then this one would be the issue:

```tsx
<label class="ods-field-label">
  My label
  <span class="ods-helper-text">Some help</span>
  <input class="ods-input" />
</label>
```

Personally, I guess the first example having 4px padding top is the least problematic, `<label>` with no copy/text is quite unusual.

packages/react/src/components/field-label/field-label.stories.tsx
```diff
@@ -0,0 +1,90 @@
...
+export const WithInput = ({
+  id,
+  label,
+  ...restFieldLabelProps
+}: FieldLabelWithInputProps) => (
+  <>
+    <FieldLabel {...restFieldLabelProps} htmlFor={id}>
+      {label}
+    </FieldLabel>
+    <Input id={id} />
```
rabelloo, line 62: I feel we should encourage nesting the input inside the label, but that's subjective.

```tsx
<FieldLabel {...restFieldLabelProps} htmlFor={id}>
  {label}
  <Input id={id} />
</FieldLabel>
```

What do you think?

packages/react/src/components/input/input.tsx
```diff
@@ -1,21 +1,39 @@
...
 import { withRef } from '../../utils';
 
+const idPrefix = 'castor_input';
+let idCount = 0;
+
 export const Input = withRef(
   (
-    { type = 'text', invalid, className, ...restProps }: InputProps,
+    {
+      id = `${idPrefix}_${idCount++}`,
```
rabelloo, line 12: 🤔    interesting decision not to go with uuidv4, mind sharing the thought process?

I guess this one prevents collisions 100%, rather than 1-in-a-million, but it's not opaque (which can be both an advantage and disadvantage).

packages/react/src/components/input/input.tsx
```diff
@@ -1,21 +1,39 @@
...
-      className={classy(c('input'), m({ invalid }), className)}
-    />
+    <FieldLabelWrapper id={id}>
+      {{
+        children,
+        element: (
+          <input
+            {...restProps}
+            ref={ref}
+            id={id}
```
rabelloo, line 28: 🤓  we include an `id` even if the input has no label.
Not a problem, just a bit unnecessary.

packages/react/src/components/textarea/textarea.tsx
```diff
@@ -1,29 +1,42 @@
...
+      {{
+        children,
+        element: (
+          <textarea
+            {...restProps}
+            ref={ref}
+            id={id}
+            rows={rows}
+            className={classy(c('textarea'), m({ invalid }), className)}
+            style={{ ...style, resize }}
```
rabelloo, line 33: Now this gets tricky again.
Should we apply `style` and `className` to the parent `<label>` like we do with Checkbox and Radio?
We should at the very least apply `data-` and `aria-` attributes, I suppose.

rabelloo (approved):

packages/react/src/components/input/input.tsx
josokinas: Regarding uuidv4 - as we do not use any libs, obvious was to have own utility for such. I was looking at best examples and still they seem too complex.

Then again I was thinking, does UUID best fits in such case? It's too universal, so having a custom ID system with "castor" prefix IMO is even nicer to look up for something in the DOM.

I did use such technique (if you can call it) previously with Redux, and it was actually working well.

packages/core/src/components/field-label/field-label.scss
josokinas: LOL, yes...

packages/react/src/components/input/input.tsx
rabelloo: Fair enough. A custom uuidv4 function is simple to execute but hard to read.
I suppose the advantages of a non-opaque id outweigh the disadvantages in this case.

Should we make it start from 1 rather than 0 though?
```suggestion
      id = `${idPrefix}_${++idCount}`,
```

packages/core/src/components/field-label/field-label.scss
josokinas: 4px at the top is by design.

Design also requires to always follow 8px grid. So this then gets a little tricky.

Your example below has 48px of input size + 4px padding of a label, then 4px margin for an input, so 56px.

```html
<label class="ods-field-label">
  <input class="ods-input" />
</label>
```

This example also adds a 24px text (but still keeps 8px grid).

```html
<label class="ods-field-label">
  My label
  <input class="ods-input" />
</label>
```

This example below also adds a 24px helper text, so in the end is 104px - still 8px grid.

```html
<label class="ods-field-label">
  My label
  <span class="ods-helper-text">Some help</span>
  <input class="ods-input" />
</label>
```

Do you see any potential issue here?

packages/react/src/components/field-label/field-label.stories.tsx
josokinas: Probably you have noticed that both Input/Textarea and now FieldLabel stories have different usage examples. I wanted to show variations of how these can be composed.

Whether they should be nested or not - I'm not too sure what's more appropriate. But with this I wanted to show that thet can be separate but _must_ be connected via the ID. For example, if you'd to have a table with two columns one which hold the label, then other holds an input.

Do you strongly suggest that these should be nested? I don't mind changing.

packages/core/src/components/field-label/field-label.scss
rabelloo: Ah right, it'll be a 8px total of empty space above the `<input />`.

I mean, you shouldn't use `<label />` if you're not going to have text inside it, right?
The empty space by spec should not be there if there's not copy/text for a label, but otherwise we complicate the usage by a lot.

I guess we just say this is incorrect usage:
```tsx
<label class="ods-field-label">
  <input class="ods-input" />
</label>
```

Either include copy/text for the label or remove the `<label />` element altogether.
Agree?

packages/react/src/components/field-label/field-label.stories.tsx
rabelloo: No, that makes sense. Showing different options is good too.

packages/react/src/components/input/input.tsx
josokinas: The problem here is that I do allow to set `id` via the prop, then default to internal one. I suppose a check could be done to e.g. see if a prefix is there in the ID, but what if someone actually decides to use such as custom one? Not very likely, but still might introduce a bug?

Obviously, instead of defaulting `id` to internal one I could set it within component in case of label is used but no `id` is set via the prop, but that also includes a little more component logic.

Would you prefer this?

packages/core/src/components/field-label/field-label.scss
josokinas: Yes.

Let's point this out to designers (if they will understand what's the issue here).

At the end there will still be additional docs provided, e.g. on how to set required/optional fields.

Would you suggest to include anything in current docs, if so what and where?

packages/react/src/components/textarea/textarea.tsx
josokinas: Here comes the annoying part... `FieldLabelWrapper` either wraps with `FieldLabel` _or_ renders input/textarea component directly.

For such case these two (`className` and `style`) should go back to input/textarea component, but then `cloneElement` should be used to pass the props. Which I know you don't like.

packages/react/src/components/input/input.tsx
josokinas: I've refactored this part now, as there was also an issue with re-renders to generate ID each time.

packages/react/src/components/input/input.tsx
```diff
@@ -1,21 +1,48 @@
...
-      className={classy(c('input'), m({ invalid }), className)}
-    />
-  )
+  ): JSX.Element => {
+    const [id, setId] = useState<string | undefined>(externalId);
+
+    useEffect(() => {
+      if (externalId || children)
+        setId(externalId || `${idPrefix}_${++idCount}`);
+    }, [externalId]);
```
rabelloo, line 26: There's a few bugs in here, and there's also an extra re-render when `externalId` changes.
We can get rid of the effect and simplify it a bit by using a state initialiser function.

```suggestion
    const [autoId, setAutoId] = useState(() => children && `${idPrefix}_${++idCount}`);
    const id = externalId || autoId;
```

packages/core/src/components/field-label/field-label.scss
rabelloo: Hmm good question. For React probably only on FieldLabel's story, as `Input` and `Textarea` should receive the label as children and not worry about `FieldLabel` itself.

For HTML stories we'll have to be more thorough.

packages/react/src/components/input/input.tsx
josokinas: Interesting, but this won't catch the `children` change would it?

packages/react/src/components/textarea/textarea.tsx
rabelloo: Argh. Or we have ternaries to apply them in either or.

Let's leave it like this for now, I suppose further control over `FieldLabel` is always possible with the alternate rendering option:
```ts
<FieldLabel data-test="foo">
  <Input data-test="bar" />
</FieldLabel>
```

josokinas: Checkbox/Radio had issue because of an indicator, and original `<input>` hidden. This should not be problematic, but can always revisit if we spot something.

packages/react/src/components/input/input.tsx
rabelloo: It wouldn't, but that's better because the id is constant regardless of how many times `children` changes.
There's no harm in applying the id if the `children` is for some reason later removed, but if you really wanted to you could with
```ts
const id = externalId || (children ? autoId : undefined);
```

packages/react/src/components/input/input.tsx
```diff
@@ -1,21 +1,46 @@
...
-    <input
-      {...restProps}
-      ref={ref}
-      type={type}
-      className={classy(c('input'), m({ invalid }), className)}
-    />
-  )
+  ): JSX.Element => {
+    const [autoId] = useState<string | undefined>(() =>
+      children ? `${idPrefix}_${++idCount}` : undefined
```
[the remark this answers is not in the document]
josokinas: Yeah, you're right.

Thought have to use `children ? autoId : undefined` otherwise:

Screenshot 2021-02-09 at 15 50 07

rabelloo:
LGTM

PR #285, opened 2021-02-09
fix: peer dependency versions

rabelloo:
## Purpose

Fix to allow newer versions too, like React 17 or castor-icons 2.

## Approach

Change the versions.

## Testing

npm should not complain about installing Castor with React 17 after merging.

## Risks

None.

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -26,8 +26,8 @@
     "!src/**/*.d.ts"
   ],
   "peerDependencies": {
-    "@onfido/castor": "1.0.0",
-    "@onfido/castor-icons": "^1.0.0",
-    "react": "^16.8.0"
+    "@onfido/castor": "^1.0.0",
+    "@onfido/castor-icons": ">=1.0.0",
+    "react": ">=16.8.0"
   }
 }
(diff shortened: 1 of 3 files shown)


packages/react/package.json
```diff
@@ -26,8 +26,8 @@
     "!src/**/*.d.ts"
   ],
   "peerDependencies": {
-    "@onfido/castor": "1.0.0",
-    "@onfido/castor-icons": "^1.0.0",
-    "react": "^16.8.0"
+    "@onfido/castor": "^1.0.0",
```
josokinas, line 29: `@onfido/castor` here is being bumped with a repository version bump, to a specific exact version.

If you think that it should use `^`, then a [version bumper](https://github.com/onfido/castor/blob/main/scripts/bump-core-peer-dependency.cjs#L16) should be modified.

If not, I'd say to revert to what it was:

```suggestion
    "@onfido/castor": "1.0.0",
```

josokinas (approved):

packages/react/package.json
rabelloo: It should probably allow newer minors and patches too, right?

For example, if you merge dependabot PRs one by one, it'll be annoying because when `castor` and `castor-react` are out of sync `npm` will complain.

I'll update the bumper if you don't mind.

PR #291, opened 2021-02-10
fix(core): make sure field-label follows 8px grid

josokinas:
## Purpose

`FieldLabel` can be used either individually, or within (and next to) `Input` and `Textarea` components. But it always should follow an 8px grid.

## Approach

Add padding for `FieldLabel` top/bottom of 2 x 4px, then use negative margin in case an input or textarea is used as children to remove the bottom padding.

## Testing

On Storybook inspecting whole component composition so it follows 8px grid.

## Risks

It assumes text is followed by input or textarea as last child.


rabelloo (approved):
Weird requirement, but looks good.

PR #292, opened 2021-02-10
docs: show how to make optional and required field-label

josokinas:
## Purpose

`FieldLabel` does not provide out-of-box solution for required and optional identification, such should be composed manually.

## Approach

Two new stories shows how "optional" and "required" field labels can be composed.

## Testing

On Storybook.

## Risks

N/A

### DIFF
--- a/packages/react/src/components/field-label/field-label.stories.tsx
+++ b/packages/react/src/components/field-label/field-label.stories.tsx
@@ -1,3 +1,4 @@
+import { color } from '@onfido/castor';
 import {
   FieldLabel,
   FieldLabelProps,
@@ -22,6 +23,32 @@ export default {
 
 export const Playground: Story<FieldLabelProps> = (props: FieldLabelProps) => (
   <FieldLabel {...props} />
+);
+
+export const AsOptional: Story<FieldLabelProps> = ({
+  children,
+  ...restProps
+}: FieldLabelProps) => (
+  <FieldLabel {...restProps}>
+    <span>
+      {children}
+      &nbsp;
+      <span style={{ color: color('content-secondary') }}>(optional)</span>
+    </span>
+  </FieldLabel>
+);
+
+export const AsRequired: Story<FieldLabelProps> = ({
+  children,
+  ...restProps
+}: FieldLabelProps) => (
+  <FieldLabel {...restProps}>
+    <span>
+      {children}
+      &nbsp;
+      <span style={{ color: color('content-negative') }}>*</span>
+    </span>
+  </FieldLabel>
 );
 
 interface FieldLabelWithHelperTextProps extends FieldLabelProps {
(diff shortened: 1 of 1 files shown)


josokinas:
As requested by @Subcide ☝️

packages/react/src/components/field-label/field-label.stories.tsx
```diff
@@ -24,6 +25,32 @@ export const Playground: Story<FieldLabelProps> = (props: FieldLabelProps) => (
...
 
+export const AsOptional: Story<FieldLabelProps> = ({
+  children,
+  ...restProps
+}: FieldLabelProps) => (
+  <FieldLabel {...restProps}>
+    <span>
+      {children}
+      &nbsp;
+      <span style={{ color: color('content-secondary') }}>(optional)</span>
```
rabelloo, line 36: 💅🏻
```suggestion
      <span style={{ color: color('content-secondary') }}> (optional)</span>
```

packages/react/src/components/field-label/field-label.stories.tsx
```diff
@@ -24,6 +25,32 @@ export const Playground: Story<FieldLabelProps> = (props: FieldLabelProps) => (
...
+
+export const AsRequired: Story<FieldLabelProps> = ({
+  children,
+  ...restProps
+}: FieldLabelProps) => (
+  <FieldLabel {...restProps}>
+    <span>
+      {children}
+      &nbsp;
+      <span style={{ color: color('content-negative') }}>*</span>
```
rabelloo, line 49: 💅🏻
```suggestion
      <span style={{ color: color('content-secondary') }}> *</span>
```

rabelloo (approved):

packages/react/src/components/field-label/field-label.stories.tsx
josokinas: I had like such originally, but then the code example becomes like such:

Screenshot 2021-02-10 at 15 34 43

So using `&nbsp;` might not look like best in code, but at least renders that code snippet better.

Although, now that I'm thinking probably it's not correct code example, so copy-pasting it would not reproduce the same output...

josokinas: The space will be missing, as React requires such either be set as `&nbsp;` or `{' '}`.

You just confused me now, shall I go with your suggestion to have a bit odd code samples but at least correct ones?

rabelloo: I'm also confused, the example does include a `{' '}` 🤔 

I'm fine with either, folk accustomed to HTML will understand the `&nbsp;` easily, React-only people might have to think a bit.
Then `{' '}` is the opposite 🤷🏻

PR #293, opened 2021-02-10
feat(core): automatically include field-label (for input/textarea)

josokinas:
## Purpose

`FieldLabel` is not included within `Input` and `Textarea` automatically when children are passed (talking about React components here), so CSS should also include styling for such automatically.

## Approach

Automatically include (only once) CSS for "field-label: component.

## Testing

Locally, on Storybook.

## Risks

CSS includes also in case label is not used.

### DIFF
--- a/packages/core/src/components/index.scss
+++ b/packages/core/src/components/index.scss
@@ -18,9 +18,12 @@
 @forward './textarea/textarea';
 
 @mixin components() {
+  // conditionally includable in other components
+  @include field-label.FieldLabel();
+
+  // all the rest
   @include button.Button();
   @include checkbox.Checkbox();
-  @include field-label.FieldLabel();
   @include helper-text.HelperText();
   @include icon.Icon();
   @include input.Input();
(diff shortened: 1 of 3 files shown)


packages/core/src/components/index.scss
```diff
@@ -20,11 +22,14 @@
...
   @include helper-text.HelperText();
   @include icon.Icon();
   @include input.Input();
   @include radio.Radio();
   @include search.Search();
   @include textarea.Textarea();
+
+  @if not includes.the('field-label') {
+    @include field-label.FieldLabel();
+  }
```
rabelloo, line 34: I suppose we could hoist the include to the top and it will always execute before `Input` and `Textarea`, thus removing the need for the `@if` here.

What do you think?

```tsx
@include field-label.FieldLabel();

@include input.Input();
@include textarea.Textarea();
```

rabelloo (approved):
A nice improvement, in my opinion a safe assumption that most Inputs and Textareas will contain a label

packages/core/src/components/index.scss
josokinas: Sure, just added little comment separating two lists.

PR #294, opened 2021-02-10
docs(react): identify that helper-text also used for field-label

josokinas:
## Purpose

`HelperText` can be used within `Checkbox`, `Radio`, `FieldLabel` components.

## Approach

Identify where `HelperText` can be used.

## Testing

N/A

## Risks

N/A

### DIFF
--- a/packages/react/src/components/helper-text/helper-text.tsx
+++ b/packages/react/src/components/helper-text/helper-text.tsx
@@ -2,7 +2,7 @@ import { c, classy, HelperTextProps as BaseProps, m } from '@onfido/castor';
 import React from 'react';
 
 /**
- * Intended to be used within `Checkbox` and `Radio` components.
+ * Intended to be used with field components, incl. `FieldLabel` itself.
  */
 export const HelperText = ({
   disabled,


packages/react/src/components/helper-text/helper-text.tsx
```diff
@@ -2,7 +2,8 @@ import { c, classy, HelperTextProps as BaseProps, m } from '@onfido/castor';
 import React from 'react';
 
 /**
- * Intended to be used within `Checkbox` and `Radio` components.
+ * Intended to be used within `Checkbox` and `Radio`, or `FieldLabel` (incl. as
+ * children of `Input` and `Textarea`) components.
```
rabelloo, line 6: This is a little bit confusing.

Does it mean this?

```tsx
<Checkbox | Radio | Input | Textarea | FieldLabel>
  <HelperText />
</>
```

josokinas: Did a little rephrase, have a look now.

rabelloo: I like the simplicity.
Doesn't mention the specific components, which for me is an advantage, but is objectively less explicit.

rabelloo (approved):

PR #296, opened 2021-02-10
docs(react): inform that form element children acts as labels

josokinas:
## Purpose

Form elements render labels when such are set via the `children` prop.

## Approach

Inform about behaviour in Storybook stories of React.

## Testing

On Storybook.

## Risks

N/A

### DIFF
--- a/packages/react/src/components/checkbox/checkbox.stories.tsx
+++ b/packages/react/src/components/checkbox/checkbox.stories.tsx
@@ -7,7 +7,10 @@ export default {
   component: Checkbox,
   argTypes: {
     ...omit<CheckboxProps>('className', 'style'),
-    children: { control: 'text' },
+    children: {
+      description: 'Acts as a label for an `<input>`.',
+      control: 'text',
+    },
     bordered: {
       table: { type: { summary: 'boolean' } },
     },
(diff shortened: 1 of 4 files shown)


packages/react/src/components/checkbox/checkbox.stories.tsx
```diff
@@ -19,6 +22,7 @@ export default {
     },
   },
   args: {
+    children: '',
```
rabelloo, line 25: 🤔  curious, would not setting anything here be any different?

rabelloo (approved):

packages/react/src/components/checkbox/checkbox.stories.tsx
josokinas: Very little - Storybook adds `-`:

Screenshot 2021-02-11 at 10 40 30

I'm not entirely sure why, but I have noticed this some time ago also with some other story.

rabelloo: Ah, it's probably because it's not sure of the type.
If we set it under `argTypes` it probably changes to `string` or whatever.
Anyway, looks good as is.

PR #298, opened 2021-02-10
fix(core): compile field-label only once

josokinas:
## Purpose

Following https://github.com/onfido/castor/pull/293 main compilation of all components is wrong, as conditional inclusion uses `includes` for keeping a map of what has been included.

## Approach

Rollback to checking for whether "field-label" was already included.

Another option is to exclude this component from a list of all components completely, however then we might introduce a bug when "field-label" is removed from both "input" and "textarea".

## Testing

Tested local build.

## Risks

N/A

### DIFF
--- a/packages/core/src/components/index.scss
+++ b/packages/core/src/components/index.scss
@@ -1,3 +1,5 @@
+@use './includes';
+
 @use './button/button';
 @forward './button/button';
 @use './checkbox/checkbox';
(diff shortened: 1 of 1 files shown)


rabelloo (approved):
Right, we forgot about this.
Unless we check whether or not it has been included, it won't be set.

Yet another option is doing something like
```scss
@mixin FieldLabel() {
  includes.the('field-label');
}
```

Which means it marks itself as included when called.

But I'm happy with the current.

josokinas:
> Which means it marks itself as included when called.

That's actually a good suggestion.

Maybe each component by default could mark itself included.

Let's see later if we have another situation of such, then we can implement something a little more smart.

PR #301, opened 2021-02-11
fix(core): correct placement of "bordered" indicator-container

josokinas:
## Purpose

Placement of "bordered" indicator-container content is wrong.

## Approach

Setting placement of content to be "start" position.

## Testing

On Storybook checking `Checkbox` and `Radio` components with labels and "bordered".

## Risks

N/A


josokinas:
Similar to @kulak-at problem ☝️

rabelloo (approved):

Issue #5, opened 2020-11-30
Validation


josokinas:
Implement the Validation component.

PR #303, opened 2021-02-11
feat: add validation component

josokinas:
## Purpose

Add Validation component.

## Approach

Implements Validation component.

This closes issue #5.

## Testing

On Storybook preview.

## Risks

N/A

### DIFF
--- a/packages/core/src/components/validation/validation.scss
+++ b/packages/core/src/components/validation/validation.scss
@@ -1,0 +1,15 @@
+@use '../../helpers';
+
+@mixin Validation() {
+  .ods-validation {
+    @include helpers.font('300-regular');
+
+    color: helpers.color('content-negative');
+    display: grid;
+    gap: helpers.space(1);
+    grid-auto-flow: column;
+    padding: helpers.space(0.5) 0;
+    place-content: start;
+    width: fit-content;
+  }
+}
--- a/packages/core/src/components/validation/validation.scss
+++ b/packages/core/src/components/validation/validation.scss
@@ -4,12 +4,15 @@
   .ods-validation {
     @include helpers.font('300-regular');
 
-    color: helpers.color('content-negative');
     display: grid;
     gap: helpers.space(1);
     grid-auto-flow: column;
     padding: helpers.space(0.5) 0;
     place-content: start;
     width: fit-content;
+
+    &.-error {
+      color: helpers.color('content-negative');
+    }
   }
 }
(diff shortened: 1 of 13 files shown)


packages/react/src/components/validation/validation.tsx
```diff
@@ -0,0 +1,30 @@
+import { c, classy, m } from '@onfido/castor';
+import { Icon } from '@onfido/castor-react';
+import React from 'react';
+
+/**
+ * Intended to be used alongside field components.
+ *
+ * `Validation` uses an `Icon` that requires `Icons` (SVG sprite) to be included
```
rabelloo, line 8: If you use `showIcon`, we should note that.

packages/react/src/components/validation/validation.tsx
```diff
@@ -0,0 +1,30 @@
...
+ */
+export const Validation = ({
+  showIcon,
+  children,
+  className,
+  ...restProps
+}: ValidationProps): JSX.Element => (
+  <div
+    {...restProps}
+    className={classy(c('validation'), m({ showIcon }), className)}
```
rabelloo, line 21: I don't think `m({ showIcon })` is doing anything, is it?

packages/react/src/components/validation/validation.tsx
```diff
@@ -0,0 +1,30 @@
...
+  >
+    {showIcon && <Icon name="error" />}
+    {children}
+  </div>
+);
+
+export type ValidationProps = JSX.IntrinsicElements['div'] & {
+  showIcon?: boolean;
```
rabelloo, line 29: 💅🏻  how do you feel about `withIcon`?

packages/core/src/components/validation/validation.scss
```diff
@@ -0,0 +1,15 @@
+@use '../../helpers';
+
+@mixin Validation() {
+  .ods-validation {
```
rabelloo, line 4: My issue with the name is that it's always a red/negative/error message, even according to its own style.

I'd call it `error` instead, or something that denotes an unsuccessful/invalid state of validation.

Like, you wouldn't use this component for any other kind of message, for example:

```tsx
<Validation success>Your password complexity is safe enough!</Validation>
```

josokinas: I understand this.

However, component naming should still align both in design and engineering. I will suggest this change to designers.

packages/react/src/components/validation/validation.tsx
josokinas: Good point, I added but realised no need to do anything with styling.

rabelloo (approved):

PR #325, opened 2021-02-17
docs: inform about using with typescript

josokinas:
## Purpose

Having `@onfido/castor-icons` as peer dependency for types might be problematic when using TypeScript + npm v6 without icons.

## Approach

Inform about situation, and provide recommendation on how to resolve the problem.

## Testing

N/A

## Risks

N/A

### DIFF
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -181,4 +181,4 @@ Or using the Sass helper:
 
 Components have basic types exported so that you could create your own JS/TS variations using any renderer chosen (for example, Vue or Angular).
 
-Please note that the `@onfido/castor-icons` package is a peer dependency and is required to be installed for types to work properly. Yarn and newer versions of npm (using lockfile v2) will resolve it, however otherwise it must be installed manually even if you don't plan to use icons.
+Please note that the `@onfido/castor-icons` package is a peer dependency and is required to be installed for **types** to work properly. Yarn and newer versions of npm (using lockfile v2) will resolve it, otherwise it must be installed manually even if you don't plan to use icons.
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -178,3 +178,7 @@ Or using the Sass helper:
 ```html
 <button class="ods-button -action--primary round">Round Button</button>
 ```
+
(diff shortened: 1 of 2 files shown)


packages/core/README.md
```diff
@@ -178,3 +178,7 @@ Or using the Sass helper:
 ```html
 <button class="ods-button -action--primary round">Round Button</button>
 ```
+
+Components have basic types exported so that you could create your own JS/TS variations using any renderer chosen (for example, Vue or Angular).
+
+Please note that the `@onfido/castor-icons` package is a peer dependency and is required to be installed for types to work properly. Yarn and newer versions of npm (using lockfile v2) will resolve it, however otherwise it must be installed manually even if you don't plan to use icons.
```
rabelloo, line 184: 💅🏻 
```suggestion
Please note that the `@onfido/castor-icons` package is a peer dependency and is required to be installed for **types** to work properly. Yarn and newer versions of npm (using lockfile v2) will resolve it, otherwise it must be installed manually even if you don't plan to use icons.
```

packages/react/README.md
```diff
@@ -58,6 +58,14 @@ You may wish to configure your bundler to transpile to a different module syntax
 
 For example, you might choose UMD module syntax targeting ES5 if your app needs to support IE11 (please note that Castor is not tested in Internet Explorer).
 
+### Use with TypeScript
+
+Components extend base prop types with JSX additions.
+
+When using with TypeScript, always import types from `@onfido/castor-react`.
+
+Please note that the `@onfido/castor-icons` package is a peer dependency and is required to be installed for types to work properly. Yarn and newer versions of npm (using lockfile v2) will resolve it, however otherwise it must be installed manually even if you don't plan to use icons.
```
rabelloo, line 67: 💅🏻 
```suggestion
Please note that the `@onfido/castor-icons` package is a peer dependency and is required to be installed for **types** to work properly. Yarn and newer versions of npm (using lockfile v2) will resolve it, otherwise it must be installed manually even if you don't plan to use icons.
```

rabelloo (approved):

PR #327, opened 2021-02-17
fix(core): correctly space full width checkbox and radio

rabelloo:
## Purpose

Fixes Checkbox inside containers that force its width to be larger than its content.

image

## Approach

Make the checkbox indicator use only its own width (`auto`) and the label use the rest of available space (`1fr`).

## Testing

Can't reproduce on Storybook 🤔 
But essentially a `display: grid` parent and a Checkbox inside.

## Risks

None.

### DIFF
--- a/packages/core/src/internal/indicator-container/indicator-container.scss
+++ b/packages/core/src/internal/indicator-container/indicator-container.scss
@@ -12,7 +12,7 @@
     color: helpers.color('content-main');
     display: inline-grid;
     gap: $padding;
-    grid-auto-flow: column;
+    grid: auto-flow / auto 1fr;
     padding: $padding;
     place-content: start;
     position: relative;


josokinas (approved):
If you need the fix, please release a patch. Otherwise let's leave for next minor release.

P.S. We usually don't scope to components whithin a commit message, maybe better scope it as other commits to package? So `fix(core): correctly space full width on checkbox and radio`, or just `fix(core): correctly space full width on input label`.

rabelloo:
Don't need it right now, but might be good to release a patch including other fixes soon.

I'll update the commit message to what you suggested when merging.

josokinas:
We have v1.1 scheduled to be released next Friday, however we can do it also earlier if all current form components are good to go.

PR #328, opened 2021-02-17
fix(core): correct "disabled" styling for checkbox and radio

josokinas:
## Purpose

Checkbox within "checked" and also "disabled" state has wrong styling:

image

## Approach

- correctly styling "disabled" checkbox state via "after" and "before" pseudo selectors
- move background color change for "disabled" indicator container to "radio" (as this is only needed by it)

## Testing

On Storybook.

## Risks

N/A

### DIFF
--- a/packages/core/src/components/checkbox/checkbox.scss
+++ b/packages/core/src/components/checkbox/checkbox.scss
@@ -74,7 +74,13 @@
   }
 
   // indicating [disabled] input
-  .ods-checkbox:disabled + .ods-input-indicator::after {
-    border-color: helpers.color('content-disabled');
+  .ods-checkbox:disabled + .ods-input-indicator {
+    &::after {
+      border-color: helpers.color('content-disabled');
+    }
+
+    &::before {
+      background-color: helpers.color('background-disabled');
+    }
   }
 }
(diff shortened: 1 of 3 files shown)


packages/core/src/components/checkbox/checkbox.scss
```diff
@@ -74,7 +74,13 @@
...
-  .ods-checkbox:disabled + .ods-input-indicator::after {
-    border-color: helpers.color('content-disabled');
+  .ods-checkbox:disabled + .ods-input-indicator {
+    &::after {
+      border-color: helpers.color('content-disabled');
+    }
+
+    &::before {
+      background-color: helpers.color('background-disabled');
+    }
```
rabelloo, line 84: I don't understand how this fixes it, seems to be the exact same rule/selector in a different file.
Is it just a specificity thing?

rabelloo (approved):

packages/core/src/components/checkbox/checkbox.scss
josokinas: It needs to come _after_ all other stylings like for "checked", as "disabled" should make the styling no matter what is applied previously.

It also _must_ be on each component styles individually, as indicator component is included first, after which follows all other additional CSS styles.

rabelloo: I see, it is specificity then. Thanks.

rabelloo:
LGTM

PR #329, opened 2021-02-17
feat(core): add .-touched:invalid as an alternative to .-invalid

rabelloo:
## Purpose

Enable upcoming `<Form>` and `<Field>` components, as well as customer specific implementations, to simply mark a field/input as `.-touched` to take full advantage of HTML5 validation.

Adding `.-touched` on blur is much simpler than replicating the entire validation logic in JavaScript.
This is even more valuable for non-React (or vanilla) customers that want minimal JS.

For example:

https://user-images.githubusercontent.com/18623773/108243003-7edc9700-7145-11eb-8cde-82f3d9976240.mov

## Approach

Add `.-touched:invalid` to every occurrence of `.-invalid`.

## Testing

Video above.

## Risks

More power means more responsibility.

### DIFF
--- a/packages/core/src/internal/indicator-container/indicator-container.scss
+++ b/packages/core/src/internal/indicator-container/indicator-container.scss
@@ -83,8 +83,8 @@
     > .ods-checkbox:checked,
     > .ods-checkbox:indeterminate,
     > .ods-radio:checked {
-      &:not(.-touched:invalid:disabled) + .ods-input-indicator,
-      &:not(.-invalid:disabled) + .ods-input-indicator {
+      &:not(.-invalid:disabled) + .ods-input-indicator,
+      &:not(.-touched:invalid:disabled) + .ods-input-indicator {
         background-color: helpers.color('background-action-subtle');
         border-color: helpers.color('border-action-subtle');
       }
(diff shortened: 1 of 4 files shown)


packages/core/src/internal/indicator-container/indicator-container.scss
```diff
@@ -82,7 +83,8 @@
     > .ods-checkbox:checked,
     > .ods-checkbox:indeterminate,
     > .ods-radio:checked {
-      &:not(:disabled):not(.-invalid) + .ods-input-indicator {
+      &:not(.-touched:invalid:disabled) + .ods-input-indicator,
+      &:not(.-invalid:disabled) + .ods-input-indicator {
```
josokinas, line 87: Minor, however to follow other changed selectors:

```suggestion
      &:not(.-invalid:disabled) + .ods-input-indicator,
      &:not(.-touched:invalid:disabled) + .ods-input-indicator {
```

josokinas (approved):
Specificity is crazy for Checkbox and Radio 🙈 But I have no other solution because of complex designs...

rabelloo:
> Specificity is crazy for Checkbox and Radio 🙈 But I have no other solution because of complex designs...

I know right.
At this point we might as well loosen the stylelint rule to make use of nesting and make it more readable.

Or we could try and simplify the design a bit, would be nice to use the `input` instead of needing the extra `span`.

josokinas:
> Or we could try and simplify the design a bit, would be nice to use the `input` instead of needing the extra `span`.

Let's suggest this to designers, and see what they say.

Otherwise LGTM, feel free to merge this.

PR #330, opened 2021-02-17
feat: support custom transition-duration

rabelloo:
## Purpose

Allow customers to determine their own transition-duration and apply it to relevant properties in components.
Also, completely disable motion when that is the browser preference (`prefers-reduced-motion`).

## Approach

Make current transitions read variable instead of fixed value.

## Testing

Storybook, should be exactly the same as before.

Then add `--ods-transition-property: 0.5s` to the story iframe's `html` and assert duration has changed.

## Risks

None?

### DIFF
--- a/packages/core/src/theme/theme.scss
+++ b/packages/core/src/theme/theme.scss
@@ -14,10 +14,10 @@
       @include castor.tokens();
     }
 
+    @content;
+    
     @media (prefers-reduced-motion) {
       --ods-transition-duration: 0s;
     }
-
-    @content;
   }
 }
--- a/packages/core/src/theme/theme.scss
+++ b/packages/core/src/theme/theme.scss
@@ -15,9 +15,9 @@
     }
 
     @content;
-  }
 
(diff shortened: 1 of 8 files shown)


rabelloo:
Cut back the feature to only current transitions as we discussed.

Also edited PR description accordingly.

Created draft PR https://github.com/onfido/castor/pull/343 for possibly reintroducing other transitions later.

josokinas (approved):
LGTM 👍

packages/core/src/theme/theme.scss
```diff
@@ -14,6 +14,10 @@
       @include castor.tokens();
     }
 
+    @media (prefers-reduced-motion) {
+      --ods-transition-duration: 0s;
+    }
+
     @content;
```
rabelloo, line 21: Almost forgot, since the variable is now set in the theme, we have to include the media query after the content to be able to replace its value.

```suggestion
    @content;
    
    @media (prefers-reduced-motion) {
      --ods-transition-duration: 0s;
    }
```

PR #333, opened 2021-02-18
feat: add fieldset component

josokinas:
## Purpose

Add Fieldset component, that is used same as `<fieldset>` element.

## Approach

Implements Fieldset component.

Later a FieldsetLegend component to be added that would make more sense.

## Testing

On Storybook preview - Checkbox and Radio stories "with validation".

## Risks

N/A

### DIFF
--- a/packages/core/src/components/fieldset/fieldset.scss
+++ b/packages/core/src/components/fieldset/fieldset.scss
@@ -1,0 +1,7 @@
+@mixin Fieldset() {
+  .ods-fieldset {
+    border: unset;
+    margin: 0;
+    padding: 0;
+  }
+}
--- a/packages/core/src/components/fieldset/fieldset.scss
+++ b/packages/core/src/components/fieldset/fieldset.scss
@@ -1,6 +1,6 @@
 @mixin Fieldset() {
   .ods-fieldset {
-    border: unset;
+    border: 0;
     margin: 0;
     padding: 0;
   }
--- a/packages/react/src/components/radio/radio.stories.tsx
+++ b/packages/react/src/components/radio/radio.stories.tsx
@@ -1,4 +1,5 @@
 import {
+  Fieldset,
   HelperText,
   Radio,
   RadioProps,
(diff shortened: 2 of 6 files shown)


packages/core/src/components/fieldset/fieldset.scss
```diff
@@ -0,0 +1,7 @@
+@mixin Fieldset() {
+  .ods-fieldset {
+    border: unset;
```
rabelloo, line 3: 💅🏻  [nitpick]

Some browsers have issues with the `unset` value.
Probably none that we care about, but still.

```suggestion
    border: 0;
```

or

```suggestion
    border: none;
```

packages/react/src/components/radio/radio.stories.tsx
```diff
@@ -89,14 +90,14 @@ export const WithValidation = ({
   withIcon,
   ...restRadioProps
 }: RadioWithValidationProps) => (
-  <>
+  <Fieldset>
```
rabelloo, line 93: This is probably better as a Field rather than a Fieldset, don't you think?

Unless this had a legend, or several Fields in it, so that you could do, for instance:

`<Fieldset disabled>`

Or is the intent to finish this story after `Legend` is done?

```ts
<Fieldset>
  <Legend />
  <Field>
     <Radio />
     <Validation />
```

rabelloo (approved):

packages/core/src/components/fieldset/fieldset.scss
josokinas: Linting does not allow "none" value for border:

Screenshot 2021-02-19 at 08 40 51

Going with "0".

packages/react/src/components/radio/radio.stories.tsx
josokinas: For us `Field` is not an option for Checkbox and Radio components - styling is not right, that is `width: min-content` breaks validation message.

Probably I will add a legend here, not sure - will see how it fits best at later PR.

rabelloo: Hmm that might be tricky if `Field` is the one holding validation context later. We'll see.

PR #338, opened 2021-02-19
feat: add fieldset-legend component

josokinas:
## Purpose

Add FieldsetLegend component, that is used same as `<legend>` element within `<fieldset>`.

## Approach

Implements FieldsetLegend component.

## Testing

On Storybook preview, including Checkbox and Radio stories "with fieldset legend".

## Risks

N/A

### DIFF
--- a/packages/core/src/components/fieldset-legend/fieldset-legend.scss
+++ b/packages/core/src/components/fieldset-legend/fieldset-legend.scss
@@ -1,0 +1,12 @@
+@use '../../helpers';
+
+@mixin FieldsetLegend() {
+  .ods-fieldset-legend {
+    @include helpers.font('300-regular');
+
+    color: helpers.color('content-main');
+    display: flex;
+    flex-flow: column wrap;
+    padding: helpers.space(0.5) 0;
+  }
+}
--- a/packages/react/src/components/checkbox/checkbox.stories.tsx
+++ b/packages/react/src/components/checkbox/checkbox.stories.tsx
@@ -73,16 +73,15 @@ interface CheckboxesWithFieldsetLegendProps extends CheckboxProps {
 }
 
 export const WithFieldsetLegend = ({
-  name,
   legend,
   ...restCheckboxProps
 }: CheckboxesWithFieldsetLegendProps) => (
   <Fieldset>
     <FieldsetLegend>{legend}</FieldsetLegend>
-    <Checkbox {...restCheckboxProps} name={name} value="1">
+    <Checkbox {...restCheckboxProps} value="1">
       one
     </Checkbox>
-    <Checkbox {...restCheckboxProps} name={name} value="2">
+    <Checkbox {...restCheckboxProps} value="2">
       two
     </Checkbox>
   </Fieldset>
--- a/packages/react/src/components/checkbox/checkbox.stories.tsx
+++ b/packages/react/src/components/checkbox/checkbox.stories.tsx
@@ -2,6 +2,7 @@ import {
   Checkbox,
   CheckboxProps,
   Fieldset,
+  FieldsetLegend,
   HelperText,
   Validation,
 } from '@onfido/castor-react';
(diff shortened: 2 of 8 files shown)


packages/core/src/components/fieldset-legend/fieldset-legend.scss
```diff
@@ -0,0 +1,12 @@
+@use '../../helpers';
+
+@mixin FieldsetLegend() {
+  .ods-fieldset-legend {
```
rabelloo, line 4: How do you feel about `FieldLegend`?

Does it make it seem like it should be a children of `Field`'s instead?

packages/react/src/components/checkbox/checkbox.stories.tsx
```diff
@@ -66,6 +67,34 @@ export const AsIndeterminate = (props: CheckboxProps) => {
...
+  name,
+  legend,
+  ...restCheckboxProps
+}: CheckboxesWithFieldsetLegendProps) => (
+  <Fieldset>
+    <FieldsetLegend>{legend}</FieldsetLegend>
+    <Checkbox {...restCheckboxProps} name={name} value="1">
+      one
+    </Checkbox>
+    <Checkbox {...restCheckboxProps} name={name} value="2">
```
rabelloo, line 85: 💅🏻  There's no need to manually destructure the name here.

```suggestion
  legend,
  ...restCheckboxProps
}: CheckboxesWithFieldsetLegendProps) => (
  <Fieldset>
    <FieldsetLegend>{legend}</FieldsetLegend>
    <Checkbox {...restCheckboxProps} value="1">
      one
    </Checkbox>
    <Checkbox {...restCheckboxProps} value="2">
```

packages/react/src/components/fieldset-legend/fieldset-legend.tsx
```diff
@@ -0,0 +1,14 @@
+import { c, classy } from '@onfido/castor';
+import React from 'react';
+
+/**
+ * Intended to be used within `Fieldset` component, providing a text legend.
```
rabelloo, line 5: 💅🏻 

```suggestion
 * Use within `Fieldset` component to provide a legend to an input.
```

packages/react/src/components/radio/radio.stories.tsx
```diff
@@ -57,6 +58,32 @@ export const Disabled = storyOf(Radio, 'disabled', [true, false], {
...
+  name,
+  legend,
+  ...restRadioProps
+}: RadiosWithFieldsetLegendProps) => (
+  <Fieldset>
+    <FieldsetLegend>{legend}</FieldsetLegend>
+    <Radio {...restRadioProps} name={name} value="yes">
+      yes
+    </Radio>
+    <Radio {...restRadioProps} name={name} value="no">
```
rabelloo, line 76: 💅🏻  same here

```suggestion
  legend,
  ...restRadioProps
}: RadiosWithFieldsetLegendProps) => (
  <Fieldset>
    <FieldsetLegend>{legend}</FieldsetLegend>
    <Radio {...restRadioProps} value="yes">
      yes
    </Radio>
    <Radio {...restRadioProps} value="no">
```

rabelloo (approved):

packages/react/src/components/fieldset-legend/fieldset-legend.stories.tsx
```diff
@@ -0,0 +1,135 @@
...
+  <Fieldset>
+    <FieldsetLegend {...restFieldsetLegendProps}>
+      {legend}
+      <HelperText>{helperText}</HelperText>
+    </FieldsetLegend>
+  </Fieldset>
+);
+WithHelperText.argTypes = omit<FieldsetLegendWithHelperTextProps>('children');
+WithHelperText.args = {
+  legend: 'Lgend',
```
rabelloo, line 81: 💅🏻 

```suggestion
  legend: 'Legend',
```

packages/core/src/components/fieldset-legend/fieldset-legend.scss
josokinas: Yes, as this is only used with `Fieldset` but should not be used with `Field`.

I think complexity will be if we want to do some functionality on `Field` that is to be also on `Fieldset`, but it's then another thought on how to share and reuse the code.

packages/react/src/components/fieldset-legend/fieldset-legend.tsx
josokinas: Will not apply this, as I'd like to avoid using "input" here.

rabelloo: No worries, maybe this then?

```suggestion
 * Use within `Fieldset` component to provide a text legend.
```

PR #377, opened 2021-02-24
docs: include theme color tokens for storybook

### LINKED ISSUE #152: Show all available colors in dropdown of icon color selection (on Storybook)
When on Storybook, a color dropdown does not include theme tokens but only regular ones.

josokinas:
## Purpose

Picker for a Castor color (currently used on an `Icon` story only) does not include theme tokens.

This closes #152.

## Approach

Merge themes, then import tokens and generate an array of color (tokens) available.

## Testing

On Storybook, `Icon` story by changing a color.

## Risks

If themes have different colors tokens set, then such would appear on a dropdown but can have no value.

This should not happen, but also something to think about on how a validation of theme tokens could be done to make sure each theme has all but not other tokens defined.

### DIFF
--- a/docs/colors/tokens.scss
+++ b/docs/colors/tokens.scss
@@ -1,0 +1,6 @@
+@use '~@onfido/castor';
+
+:export {
+  @include castor.day();
+  @include castor.night();
+}
--- a/docs/colors/colors.ts
+++ b/docs/colors/colors.ts
@@ -1,0 +1,7 @@
+import { Color } from '@onfido/castor';
+import tokens from './tokens.scss';
+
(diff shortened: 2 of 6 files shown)


docs/colors/tokens.scss
```diff
@@ -0,0 +1,6 @@
+@use '~@onfido/castor';
+
+:export {
+  @include castor.day();
+  @include castor.night();
```
rabelloo, line 5: 🤓  [no action required]

I guess merging is fine, but so would be only including one, assuming day and night always have the same structure and only have values changed.

rabelloo (approved):

PR #381, opened 2021-02-25
build: generate bundle for prototyping

josokinas:
## Purpose

A bundle for quick prototype is needed.

## Approach

Using esbuild to generate IIFE prototype bundle, that is distributed together with React package.

CDN is then used to access that prototype directly from npm registry (adding `.min` as it [supports minification](https://www.jsdelivr.com/) itself).

Also suggested approach is to use EcmaScript modules directly in the browser with Skypack.

## Testing

This is based on assumption, I can't check this 100% until a new version (expecting this to be in v1.2) is released.

## Risks

Skypack won't work for any earlier versions.


.browserslistrc
```diff
@@ -0,0 +1,5 @@
+last 2 chrome versions
+last 2 firefox versions
+last 2 opera versions
+last 2 safari versions
+not dead
```
rabelloo, line 5: We're missing several important ones here, namely Edge and the mobile (android/ios) versions of the ones listed.

I'd recommend using the defaults, then dropping IE which is 💀  and possibly other troublesome browsers.

```suggestion
defaults
not ie > 0
```

package.json
```diff
@@ -20,6 +20,9 @@
     "build:core:css": "sass packages/core/style:packages/core/dist --no-source-map -s compressed",
     "build:core:esm": "tsc -p packages/core/tsconfig.json",
     "build:react:esm": "tsc -p packages/react/tsconfig.json",
+    "preprototype": "shx rm -rf packages/*/prototype",
+    "prototype": "concurrently yarn:prototype:*",
+    "prototype:react": "parcel build packages/react/src/index.ts --out-dir packages/react/prototype --out-file react.js --global Castor --no-minify --no-source-maps",
```
rabelloo, line 25: 🤔  why no source maps? Should help troubleshoot any potential error, if one is thrown.

package.json
```diff
@@ -54,6 +57,7 @@
     "jest": "^26.6.3",
     "lint-staged": "^10.5.4",
     "optimist": "^0.6.1",
+    "parcel": "^1.12.4",
```
rabelloo, line 60: 💅🏻  We could use `esbuild` instead, should be faster too.

packages/react/README.md
```diff
@@ -81,3 +81,32 @@ const RoundButton = styled(Button)({
...
+<script src="https://cdn.jsdelivr.net/npm/@onfido/castor-react/prototype/react.min.js"></script>
+```
+
+You can also use any previously released version (v1.2.0 and up), for example:
+
+```html
+<script src="https://cdn.jsdelivr.net/npm/@onfido/castor-react@1.2.0/prototype/react.min.js"></script>
+```
+
+The same prototype file (not minified) is localed on each npm package version.
```
rabelloo, line 99:
```suggestion
The same prototype file (not minified) is located on each npm package version.
```

packages/react/README.md
```diff
@@ -81,3 +81,32 @@ const RoundButton = styled(Button)({
...
+```html
+<script type="module">
+  import { borderRadius } from 'https://cdn.skypack.dev/@onfido/castor';
+  import { Button } from 'https://cdn.skypack.dev/@onfido/castor-react';
+
+  // all other app code
+</script>
+```
+
+Also it is easy to use Skypack distributions [with Deno](https://docs.skypack.dev/skypack-cdn/code/deno).
```
rabelloo, line 112: 👍🏻

packages/react/README.md
```diff
@@ -81,3 +81,32 @@ const RoundButton = styled(Button)({
   borderRadius: borderRadius('full'),
 });
 ```
+
+## Use for prototyping
+
+If you need to prototype with Castor, the quickest way is to use the generated single file global `Castor` exposed as UMD module:
+
+```html
+<script src="https://cdn.jsdelivr.net/npm/@onfido/castor-react/prototype/react.min.js"></script>
```
rabelloo, line 90: How do you feel about just naming the file `prototype.js` and putting it in the root of each package?

`@onfido/castor-react/prototype.min.js`

That way the import doesn't include the library name twice, `react` in this case.

rabelloo (approved):

package.json
josokinas: Swithed to esbuild, just to note that ES6 it's the earliest it can compile to, but I think we're fine with support:

- Chrome 49+
- Safari 10.1+
- Firefox 45+
- Edge 14+

package.json
```diff
@@ -20,6 +20,9 @@
     "build:core:css": "sass packages/core/style:packages/core/dist --no-source-map -s compressed",
     "build:core:esm": "tsc -p packages/core/tsconfig.json",
     "build:react:esm": "tsc -p packages/react/tsconfig.json",
+    "preprototype": "shx rm -f packages/*/prototype.js*",
+    "prototype": "concurrently yarn:prototype:*",
+    "prototype:react": "esbuild packages/react/src/index.ts --outfile=packages/react/prototype.js --bundle --define:process.env.NODE_ENV=\\\"production\\\" --target=es2015 --sourcemap --global-name=Castor",
```
rabelloo, line 25: 💅🏻  I believe `NODE_ENV` can be defined as an environment variable and it will be loaded with dotenv, but I might be wrong

```suggestion
    "prototype:react": "NODE_ENV=production; esbuild packages/react/src/index.ts --outfile=packages/react/prototype.js --bundle --target=es2015 --sourcemap --global-name=Castor",
```

.gitignore
```diff
@@ -24,4 +24,5 @@ package-lock.json
 .castor*
 coverage
 dist
+packages/*/prototype.js*
```
rabelloo, line 27: 🤔  why `js*`? To support `jsx` just in case?

rabelloo (approved):

package.json
josokinas: I actually followed the official esbuild instructions: https://esbuild.github.io/getting-started/#bundling-for-the-browser I think better to leave as they suggest, easier later to debug.

.gitignore
josokinas: No, to support `.js` and also `.js.map` 😏

PR #383, opened 2021-02-26
fix(core): correct background color of simulated input

josokinas:
## Purpose

Change current "transparent" background color of simulated input to "background-input".

## Approach

Add background color of "background-input" to simulated input.

## Testing

On Storybook.

## Risks

N/A


rabelloo (approved):

rabelloo (approved):

josokinas:
Twice is better then none 👍 🙈

rabelloo:
No idea why it happened, I probably hit the button twice by mistake 😅

Issue #67, opened 2020-12-08
Spinner


coreyhume:
Implement the Loading Spinner component.

josokinas:
Should probably be named just `Spinner` https://www.uiguideline.com/components/spinner

PR #402, opened 2021-03-03
feat: add spinner component

josokinas:
## Purpose

Add Spinner component.

## Approach

Implements Spinner  component.

This closes issue #67.

## Testing

On Storybook preview.

## Risks

N/A


packages/react/src/components/spinner/spinner.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+const LabelWrapper = ({
+  children: { element, children },
+}: LabelWrapperProps): JSX.Element => {
+  if (!children) return element;
+
+  return (
+    <div className={classy(c('spinner-container'))}>
+      {element}
+      {children && <div className={classy(c('spinner-label'))}>{children}</div>}
```
rabelloo, line 33: This check is not needed, right?
`if (!children)` above already covers that branch.

```suggestion
      <div className={classy(c('spinner-label'))}>{children}</div>
```

packages/core/src/components/spinner/spinner.scss
```diff
@@ -0,0 +1,70 @@
...
+
+@mixin Spinner() {
+  @if not includes.the('spin-animation') {
+    @include animations.Spin();
+  }
+
+  .ods-spinner {
+    animation: ods-spin 0.6s linear infinite;
+    border: 0 solid helpers.color('background-spinner');
+    border-left-color: helpers.color('content-spinner');
```
rabelloo, line 14: This is real clever, not needing two elements since the filled area is 1/4 of the total.

However this won't work for determinate states.
Should I just discount it as future work?
We won't be able to release this before then, since it'll most likely be a breaking change.

packages/core/src/components/spinner/spinner.scss
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    &.-small {
+      @include mixins.square(helpers.space(2));
+
+      border-width: helpers.space(0.25);
+    }
+  }
+
+  .ods-spinner-container {
```
rabelloo, line 36: I don't think the `container` or `label` are needed.

Wouldn't this work?

```scss
.ods-spinner { // as container
  color: secondary; // will only affect text anyway
  display: grid;
  // ... all other necessary positioning styles

  &.-large {
    gap: space(0.5); // replaces margin-top/left
  }

  &:before { // as spinner
    @include animation.Spin();
  }  
}
```

packages/core/src/components/spinner/spinner.scss
josokinas: This will have to be released next week.

I'm fine with a breaking change.

The "determinate" component is not finished, I don't really know what and if it will even be needed.

TBH it's similar as also this component, sort of developing without having 100% design handover.

What would you suggest then?

rabelloo: It'll have to be two elements.

Like my other comment below, I'd recommend using pseudo elements.

```scss
.ods-spinner { // container
  &:before { // spinner background }
  &:after { // spinner fill/content }
}
```

packages/core/src/components/spinner/spinner.scss
josokinas: You're right, I managed to get rid of all of it!

packages/core/src/components/spinner/spinner.scss
josokinas: Thanks, I did follow your recommendation.

Though I still blended in both "background" and "content" into one border of a pseudo element.

rabelloo (approved):
Looks good.

If we need to support determinate states later we can add an `::after` element without having to change the API.

josokinas:
> If we need to support determinate states later...

I have consulted with a designer - this component will not have a determinate state.

Issue #3, opened 2020-11-30
Dropdown


josokinas:
Implement the Dropdown component.

rabelloo:
Should probably be called `Select` https://www.uiguideline.com/components/select-dropdown

josokinas:
> Should probably be called `Select` https://www.uiguideline.com/components/select-dropdown

I suspect both is allowed here.

rabelloo:
Maybe, but still it's a better match for the HTML element, like `input` vs `textbox`
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select

josokinas:
Release only a native version of a component, custom one coming in next Q.

PR #403, opened 2021-03-03
feat: add select component

josokinas:
## Purpose

Add Select component.

## Approach

Implements Select  component.

This closes issue #3.

## Testing

On Storybook preview.

## Risks

N/A


rabelloo:
Is the idea to release native Select before the custom?

josokinas:
No, it will still belong (native + custom) to [1.4.0](https://github.com/onfido/castor/milestone/5) release, however made this into a separate PR just to have a more clear idea on where our base is and where the custom implementation goes towards.

packages/core/src/components/select/select.scss
```diff
@@ -0,0 +1,92 @@
...
+    &.-empty {
+      color: helpers.color('content-placeholder');
+    }
+
+    + .ods-icon {
+      color: helpers.color('content-main');
+      pointer-events: none;
+      position: absolute;
+      right: helpers.space(2);
+    }
```
rabelloo, line 41: Wouldn't this icon be the same in the non-native version?

packages/core/src/components/select/select.scss
```diff
@@ -0,0 +1,92 @@
...
+
+    &.-invalid,
+    &.-touched:invalid,
+    &.-invalid + .ods-icon,
+    &.-touched:invalid + .ods-icon {
+      color: helpers.color('content-negative');
+    }
```
rabelloo, line 48: 💅🏻 [nitpick]
```suggestion
    &.-invalid,
    &.-touched:invalid {
      &,
      + .ods-icon {
        color: helpers.color('content-negative');
      }
    }
```

packages/core/src/components/select/select.scss
josokinas: Yes, but ignore non-native or "custom" at all for this.

packages/core/src/components/select/select.scss
josokinas: Can't do because of linting 🙄

packages/core/src/components/select/select.scss
```diff
@@ -0,0 +1,92 @@
...
+    @error '$border-width must be a number';
+  }
+
+  $padding-horizontal: helpers.space(1.5) - $border-width;
+  $padding-vertical: helpers.space(2) - $border-width;
+
+  padding-bottom: $padding-horizontal;
+  padding-left: $padding-vertical;
+  padding-right: $padding-vertical + helpers.space(3.5); // extra space for icon
+  padding-top: $padding-horizontal;
```
rabelloo, line 91: 💅🏻 [nitpick]

```suggestion
  padding: $padding-vertical $padding-horizontal;
  padding-right: $padding-vertical + helpers.space(3.5); // extra space for icon
```

packages/core/src/components/select/select.stories.ts
```diff
@@ -0,0 +1,110 @@
...
+Borderless.argTypes = omit<SelectProps>('borderless');
+
+export const Invalid = htmlMatrix(Select, { invalid });
+Invalid.argTypes = omit<SelectProps>('invalid');
+
+export const Disabled = htmlMatrix(Select, { disabled });
+Disabled.argTypes = omit<SelectProps>('disabled');
+
+export const WithPlaceholder = ({ placeholder, ...props }: SelectProps) =>
+  Select({ ...props, placeholder });
```
rabelloo, line 55: 💅🏻 [nitpick]

```suggestion
export const WithPlaceholder: Story<SelectProps> = (props) => Select(props);
```

packages/core/src/components/select/select.story.tsx
```diff
@@ -0,0 +1,54 @@
...
+    class: classy(c('select'), m({ borderless })),
+    children: [
+      html('select', {
+        ...props,
+        class: classy(
+          c('select-native'),
+          m({
+            borderless,
+            invalid,
+            ...(placeholder && (!value || value === '') && { empty: true }),
```
rabelloo, line 37: 💅🏻 [nitpick]

```suggestion
            ...(placeholder && !value && { empty: true }),
```

packages/core/src/components/select/select.story.tsx
```diff
@@ -0,0 +1,54 @@
...
+          })
+        ),
+        children: [
+          placeholder &&
+            html('option', {
+              value: '',
+              selected: !value || value === '',
```
rabelloo, line 44: 💅🏻 [nitpick]

```suggestion
              selected: !value,
```

packages/core/src/components/select/select.story.tsx
```diff
@@ -0,0 +1,54 @@
...
+              disabled: true,
```
rabelloo, line 45: Should probably be `hidden` instead of `disabled`, arguably.
Behaves more like a Placeholder I suppose.

```suggestion
              hidden: true,
```

packages/react/src/components/select/select.stories.tsx
```diff
@@ -0,0 +1,134 @@
...
+    },
+    disabled: {
+      table: { type: { summary: 'boolean' } },
+    },
+    invalid: {
+      table: { type: { summary: 'boolean' } },
+    },
+    placeholder: {
+      table: { type: { summary: 'string' } },
+    },
```
rabelloo, line 35: 🤔  I'm surprised these can't be picked up by `storybook-react-props` or whatever that lib is called.

packages/react/src/components/select/select.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+    return (
+      <div className={classy(c('select'), m({ borderless }))}>
+        <select
+          disabled={disabled} // will be overriden by props if set
+          {...restProps}
+          ref={ref}
+          id={id}
+          defaultValue={
+            defaultValue || value != null ? undefined : placeholder && ''
```
rabelloo, line 41: I think this is missing a parens, otherwise `defaultValue` would never apply.

```suggestion
            defaultValue || (value != null ? undefined : placeholder && '')
```

rabelloo: Actually this one might read slightly better even

```suggestion
            defaultValue || (value == null && placeholder ? '' : undefined)
```

packages/react/src/components/select/select.tsx
```diff
@@ -0,0 +1,68 @@
...
+          }
+          value={value}
+          className={classy(
+            c('select-native'),
+            m({ empty, borderless, invalid, touched }),
+            className
+          )}
+          onChange={(ev) => {
+            setEmpty(value != null ? !value : !ev.currentTarget.value);
```
rabelloo, line 50: This might be a problem if there's an option with a falsy value, like `<option value={0}>Zero<option>`.
It will think it's empty and apply placeholder styles, but it's not empty and it shouldn't.

```suggestion
            setEmpty(value != null ? !value : ev.currentTarget.value == null);
```

packages/core/src/components/select/select.scss
rabelloo: Cool. I was thinking like `.ods-select > .ods-icon` but no need to change anything now.

packages/core/src/components/select/select.scss
rabelloo: Isn't it done in the very next rule? 🤔 

```scss
&:disabled {
      &,
      + .ods-icon {
        color: helpers.color('content-disabled');
      }
    }
```

packages/react/src/components/select/select.tsx
josokinas: `ev.currentTarget.value` will always be a string, even in case of setting `value={0}` (or any other number).

packages/react/src/components/select/select.stories.tsx
josokinas: It does not pick the "type" for the table.

I believe this is because we use a `withRef`.

packages/core/src/components/select/select.story.tsx
josokinas: Why `hidden`?

I want to replicate the React version.

Also, it should still be in the list.

packages/core/src/components/select/select.scss
josokinas: No, specificity in this one is more than in "disabled" - it's `.-touched` with `:invalid` where as here it's only a `:disabled`.

rabelloo: Weird that it's treated differently 🤔 
Specificity doesn't change between both versions, only nesting does.
Oh well 🤷🏻‍♂️

packages/core/src/components/select/select.story.tsx
rabelloo: I'd assume both versions would see the change.
The arguable part is that the "placeholder option" should not appear in the dropdown list.
If you're sure we want it in the list then `disabled` is good though.

packages/react/src/components/select/select.stories.tsx
rabelloo: I wonder if `forwardRef` can be picked 🤔 
Something to check later

packages/react/src/components/select/select.tsx
rabelloo: Ah you're right, nice one.

What about this?

```tsx
<Select placeholder="foo">
  <option value="">None</option>
  <option value="a">A</option>
  <option value="b">B</option>
</Select>
```

Is there any way we could prevent `None` to be styled like a placeholder?
Or do we simply require values for every option?

```tsx
  <option value="none">None</option>
```

packages/core/src/components/select/select.scss
josokinas: Yeah, it's a linting thing, we might reconsider increasing it.

packages/core/src/components/select/select.story.tsx
josokinas: OK, I see what you mean.

I personally would be pro showing it, because otherwise when open you don't see a placeholder anymore. However I will consult with a designer.

packages/react/src/components/select/select.tsx
josokinas: Option of "None" will be styled same as placeholder, yes. However, it will not be disabled.

I guess the styling is "empty", which is correct - an empty value is being selected.

rabelloo: Fair enough, I'm fine with that.
Let's document that `value=""` will always be styled like a placeholder/empty though.

josokinas: Done 👍

rabelloo:
This is really weird
image

I wonder if should encourage a default option when there's no placeholder.

We should ask @Subcide about this too.

josokinas:
We will discuss this tomorrow.

I think weirdest one is "borderless".

However, for HTML + CSS composition this is a totally valid one.

josokinas:
@rabelloo changed all this now to what we have discussed with designer.

rabelloo:
LGTM

rabelloo (approved):

PR #411, opened 2021-03-03
fix(react): exclude "children" from search component props type

josokinas:
## Purpose

Search component uses Input component under the hood, it also passes all the rest of the props to it.

Providing "children" to Input component renders a field label alongside an input.

This however breaks the Search component if allowed:

Screenshot 2021-03-03 at 16 35 20

## Approach

Exclude "children" from Search component props type.

## Testing

Locally, when coding.

## Risks

It does not disallow to still set children, e.g. if TS is not used.

However, designs point out that field labels should not even be used on Search component.

### DIFF
--- a/packages/react/src/components/search/search.tsx
+++ b/packages/react/src/components/search/search.tsx
@@ -23,4 +23,4 @@ export const Search = withRef(
 );
 Search.displayName = 'Search';
 
-export type SearchProps = Omit<InputProps, 'invalid' | 'type'>;
+export type SearchProps = Omit<InputProps, 'type' | 'invalid' | 'children'>;


rabelloo (changes requested):
It might be good to destructure `children` and not pass it to Input then, so that JS users can't mistakenly break it.

```ts
export const Search = ({ children, ...restProps }: SearchProps) =>
    <Input {...restProps} />
```

rabelloo:
As discussed on Slack, nevermind JS users, the incurred runtime cost is probably not worth it when they have documentation and TS types available.

Might be worth omitting validation related properties from the prop types too though, for example `required`, `minlength`, `maxlength`, `min`, `max`, `pattern`

rabelloo (approved):

PR #415, opened 2021-03-04
fix(core): correct style of "invalid" input's placeholder

josokinas:
## Purpose

"invalid" Input component's placeholder should be styled as "red", as seen in design:

Screenshot 2021-03-04 at 11 46 31

Same does applie to Textarea too.

## Approach

Set color of a placeholder in "invalid" state to inherit "red" color.

## Testing

On Storybook.

## Risks

N/A


josokinas:
As discussed with @Subcide, no such change ("fix") is needed.

The placeholder always stays "grey" no matter what state the text input is in.

PR #419, opened 2021-03-05
docs: add html input stories

josokinas:
## Purpose

## Approach

Adding stories to Input component.

Please note this excludes stories that include FieldLabel and other components, as they're to come later with their stories.

## Testing

On Storybook.

## Risks

N/A

### DIFF
--- a/packages/core/src/components/input/input.stories.ts
+++ b/packages/core/src/components/input/input.stories.ts
@@ -1,0 +1,67 @@
+import { htmlMatrix, Meta, omit, Story } from '../../../../../docs';
+import { Input, InputProps } from './input.story';
+
+const disabled = [true, false] as const;
+const invalid = [true, false] as const;
+
+const type = [
+  'text',
+  'number',
+  'email',
+  'password',
+  'tel',
+  'url',
+  'search',
+] as const;
+type.toString = () => type.map((value) => `"${value}"`).join('|');
+
+export default {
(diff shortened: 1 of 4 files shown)


packages/core/src/components/input/input.stories.ts
```diff
@@ -0,0 +1,67 @@
...
+  argTypes: {
+    disabled: {
+      table: { type: { summary: 'boolean' } },
+    },
+    invalid: {
+      table: { type: { summary: 'boolean' } },
+    },
+    placeholder: {
+      table: { type: { summary: 'string' } },
+    },
```
rabelloo, line 30: 🤓  [no action]
I'm still negatively surprised `docgen` can't pick any of these up :/

rabelloo (approved):

rabelloo:
Caught a bug in the `html` helper, unrelated to this PR

image

Will open an issue

packages/core/src/components/input/input.stories.ts
josokinas: It can, but not fully.

For example, it does not indicate that the prop is a "string":

Screenshot 2021-03-05 at 12 41 11

Which I dislike...

Also, otherwise not possible to achieve alphabetical order.

rabelloo: Yea we have to figure that out at some point.

PR #420, opened 2021-03-05
docs: add html textarea stories

josokinas:
## Purpose

## Approach

Adding stories to Textarea component.

Please note this excludes stories that include FieldLabel and other components, as they're to come later with their stories.

## Testing

On Storybook.

## Risks

N/A

### DIFF
--- a/packages/core/src/components/textarea/textarea.story.tsx
+++ b/packages/core/src/components/textarea/textarea.story.tsx
@@ -1,0 +1,21 @@
+import { TextareaProps as BaseProps } from '@onfido/castor';
+import { html } from '../../../../../docs';
+import { c, classy, m } from '../../utils';
+
+export interface TextareaProps extends BaseProps {
+  rows?: number;
+  value?: string;
+}
+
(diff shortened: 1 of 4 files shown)


rabelloo (approved):

packages/core/src/components/textarea/textarea.story.tsx
```diff
@@ -0,0 +1,21 @@
...
+export const Textarea = ({
+  resize = 'vertical',
+  rows = 3,
+  invalid,
+  ...props
+}: TextareaProps) =>
+  html('textarea', {
+    ...props,
+    class: classy(c('textarea'), m({ invalid })),
+    resize,
```
rabelloo, line 19: Currently does nothing, needs to be `style`

josokinas: Good catch, a little too late... Will open a fix.

rabelloo:
My bad, left the comment after approving, just before the merge.

We can fix it on another PR I suppose.

PR #424, opened 2021-03-05
docs: correct html textarea "value"

josokinas:
## Purpose

Core and React stories should match, setting "value" is not needed however because of HTML story type it was added to the story table.

## Approach

Omit "value" property for story.

## Testing

On Storybook.

## Risks

N/A


rabelloo (approved):

josokinas:
No longer relevant because of https://github.com/onfido/castor/pull/425.

PR #439, opened 2021-03-09
feat: form

rabelloo:
## Purpose

Implements a `Form` component and related integrations and patterns to facilitate HTML5 validation with `castor-react`.

## Approach

Use Context and hooks to share state between `Form`, `Field`, `Fieldset`, `Validation` and input components.

## Testing

`form.story.tsx`

## Risks

Although the `.-touched` pattern eases CSS use a little bit, non-React users still don't get anything as nice.


packages/react/src/components/form/form.stories.tsx
```diff
@@ -0,0 +1,127 @@
...
+          <Input type="number" name="month" placeholder="MM" min={1} max={12} />
+          <Input
+            type="number"
+            name="year"
+            placeholder="YYYY"
+            min={new Date().getUTCFullYear() - 150}
+            max={new Date().getUTCFullYear()}
+          />
+        </div>
+      </FieldLabel>
```
josokinas: FieldLabel is not designed to have Input or Textarea nested also within a `<div>` (or anything else).

You end up not having a required space between label text and input(s).

For such you instead should do like this:

```suggestion
      <FieldLabel htmlFor="dob-day">
        Date of birth
        <HelperText>DD / MM / YYYY</HelperText>
      </FieldLabel>
      <div
          style={{
            display: 'grid',
            gap: '2rem',
            grid: '1fr / 5rem 5rem 6rem',
          }}
        >
          <Input type="number" id="dob-day" name="day" placeholder="DD" min={1} max={31} />
          <Input type="number" name="month" placeholder="MM" min={1} max={12} />
          <Input
            type="number"
            name="year"
            placeholder="YYYY"
            min={new Date().getUTCFullYear() - 150}
            max={new Date().getUTCFullYear()}
          />
        </div>
```

rabelloo: This is a tricky one altogether, because the both `Field` and the label can't work for all inputs at the same time.

We have to think this pattern through, but for now your suggestion makes sense, presumably you start by typing the day.

packages/react/src/components/index.ts
```diff
@@ -1,10 +1,11 @@
 export * from './asterisk/asterisk';
 export * from './button/button';
 export * from './checkbox/checkbox';
+export * from './field';
 export * from './field-label/field-label';
-export * from './field/field';
```
josokinas, line 5: How did you get this change?

I just checked locally, and to me it auto-formats as it was previously.

packages/core/src/components/validation/validation.ts
```diff
@@ -1,3 +1,4 @@
 export interface ValidationProps {
+  for?: keyof ValidityState;
   state: 'error';
```
josokinas, line 3: [No action required] This now seems little odd - `for` is validity state, but then there's another `state` that's an error. Maybe prop should be named differently, like `variant`.

packages/core/src/mixins/as-text-input.scss
```diff
@@ -12,7 +12,7 @@
   box-sizing: border-box;
   color: helpers.color('content-main');
   padding: helpers.space(1.5) - $border-width helpers.space(2) - $border-width;
-  width: helpers.space(50);
+  width: 100%;
```
josokinas, line 15: [No action required] Having a déjà vu, I though we already made this change 😅

packages/react/src/components/checkbox/checkbox.tsx
```diff
@@ -28,10 +30,11 @@ export const Checkbox = withRef(
           children,
           input: (
             <input
+              disabled={disabled} // will be overriden by props if set
```
josokinas, line 33: Do you think it is not odd to use a Field but then also have ability to override a `disabled`?

Wouldn't it be better id `disabled` takes Field as priority, defaulting to a prop?

Curious to find out what was your thinking here?

packages/react/src/components/field/field.tsx
```diff
@@ -1,12 +1,61 @@
...
+  const readInput = (event: SyntheticEvent<HTMLDivElement>) => {
+    const { disabled, validity } = event.target as HTMLInputElement;
+    update({ disabled, validity });
+  };
+
+  return (
+    <FieldProvider value={field}>
+      <div
+        {...restProps}
+        ref={ref}
```
josokinas, line 41: Doesn't this disallow to use `ref` on Field component now?

packages/react/src/components/field/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './field';
```
josokinas, line 1: So far we dont have any indexes on component level.

I would suggest to either remove this and follow what was done on the IndicatorContainer internal component (so `useField` is exported via `field.tsx`. This would be my preference, because most likely when using Field you will also use `useField`.

If you disagree, then maybe worth removing a component export here, so index would only export any additions. And then add index onto `internal/indicator-container` also.

packages/react/src/components/form/form.stories.tsx
```diff
@@ -0,0 +1,134 @@
...
+          type="number"
+          name="year"
+          placeholder="YYYY"
+          min={new Date().getUTCFullYear() - 150}
+          max={new Date().getUTCFullYear()}
+        />
+      </div>
+    </Field>
+
+    <Field>
```
josokinas, line 96: Why not using Fieldset and FieldsetLegend here?

packages/react/src/components/form/form.stories.tsx
```diff
@@ -0,0 +1,134 @@
...
+      <Checkbox name="agreement" required>
+        I agree to the <a>Privacy policy</a> and <a>Terms and conditions</a>.
+      </Checkbox>
+      <Validation state="error" for="valid">
+        Please confirm you agree
+      </Validation>
+    </Field>
+
+    <Button>Send email</Button>
```
josokinas, line 132: One thing to note, that now disabling the form does not disable the button for submission.

Would you expect such behaviour?

packages/react/src/components/form/form.tsx
```diff
@@ -0,0 +1,63 @@
...
+}
+
+export interface FormProps<T extends Values>
+  extends BaseProps,
+    Omit<JSX.IntrinsicElements['form'], 'onChange' | 'onSubmit'> {
+  onChange?: (event: FormEvent<HTMLFormElement>, values: T) => void;
+  onSubmit?: (event: FormEvent<HTMLFormElement>, values: T) => void;
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
josokinas, line 62: Curious, does this get now a little annoying? Should we better disable the rule then?

packages/react/src/components/form/getFormValues.ts
```diff
@@ -0,0 +1,24 @@
+/**
+ * Gets the values of a form element in object format, based on names.
+ * @param form Form element to extract values from.
+ * @example
```
josokinas, line 4:
```suggestion
 * Gets the values of a form element in object format, based on names.

 * @param form Form element to extract values from.

 * @example
```

josokinas (changes requested):

packages/react/src/components/validation/validation.tsx
```diff
@@ -10,18 +10,35 @@ import React from 'react';
  *
  * https://github.com/onfido/castor-icons#use-with-plain-code
  */
-export const Validation = ({
-  state,
-  withIcon,
+export function Validation({
```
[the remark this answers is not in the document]
rabelloo: It doesn't, we can go back to `const`, it just makes the syntax slightly cleaner

```ts
export const Validation = () => {}

// vs

export function Validation() {}
```

packages/core/src/components/validation/validation.ts
rabelloo: Well the idea is for the API to read like `<Validation state="error" for="valueMissing" />`, the `ValidityState` is just the object that holds which validation error an element has.

We could change the name for sure, `variant` does not convey the intent though.

Maybe `when`? Or `if`?

packages/react/src/components/checkbox/checkbox.tsx
rabelloo: Hmm perhaps `Field` doesn't need disabled at all, you should only disable an input directly, a fieldset or the whole form.

I was thinking something like this would need support
```tsx
<Field disabled>
  <Radio name="foo" disabled={false}>One</Radio>
  <Radio name="foo">Two</Radio>
</Field>
```

But perhaps you should use Fieldset for that? Dunno, thoughts?

packages/react/src/components/field/field.tsx
rabelloo: It wasn't being `forwardRef`'d before, so no change here.

However this is completely unused, nice thing you asked, we can just remove it.

packages/react/src/components/field/index.ts
rabelloo: Fair enough, I don't mind removing the index, your first suggestion.

packages/react/src/components/form/form.stories.tsx
rabelloo: I need the `Field` for `Validation`, don't really need the `Fieldset` but `FieldsetLegend` might be a good call.

Should we use `FieldsetLegend` without a `Fieldset`?

Using everything might be convoluted.

packages/react/src/components/form/form.stories.tsx
rabelloo: Yes, disabled buttons are an anti-pattern, you should be able to try to submit the form, but get validation messages that prevent you. Just like pressing enter in any field.

We can also ask @Subcide for his opinion here, but I remember him being against disabled buttons.

packages/react/src/components/validation/validation.tsx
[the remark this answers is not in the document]
josokinas: Right, so I don' mind one or another, just not a fan of having both in various places.

What do you think about refactoring all components then to be `function`?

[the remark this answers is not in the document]
rabelloo: Well some are cleaner as `const`

```ts
export const Foo = () => <div />;

// vs

export function Foo() {
  return <div />;
}
```

I wouldn't mind having both, as the best for each instance varies depending on their contents.

If you want them to be consistent, which I don't mind, I can make these `const` again.

In my opinion the worst offender is definitely the last example in this comment, just above.

packages/react/src/components/index.ts
josokinas: I guess you still have the editor one enabled? Can you try to disable and see what result it does? Otherwise we will end up changing this with our each PR.

rabelloo: Not ordering it is a Prettier offence

image

I can update our `.vscode/settings.json` with autoformat enabled, if you want, to make sure both our editors behave the same?

packages/core/src/components/validation/validation.ts
rabelloo: Oh boy, that's a breaking change. I see your point though.

In hindsight `variant` or `kind` would be better for sure, aligns with Button.

packages/react/src/components/validation/validation.tsx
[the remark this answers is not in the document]
josokinas: I don't mind the last example TBH, with explicit `return` - why not?

But anyways, then I would suggest to use `const` everywhere at least for now, maybe later we can decide on something and quickly refactor.

packages/react/src/components/checkbox/checkbox.tsx
josokinas: Yes exacly, already pointed out on a story that probably a Fieldset should be used.

Having a Fieldset to have power to disable all childre inputs is not bad 👍

packages/react/src/components/validation/validation.tsx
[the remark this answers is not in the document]
rabelloo: I guess verbosity is the only downside. The upside is it doesn't require a function body introduction when hooks are needed, for example.

Sounds good, will revert to `const`.

packages/react/src/components/form/form.stories.tsx
josokinas: Why don't you need a Fieldset?

I think this is what should surround grouped radios (and also checkboxes).

Also, as discussed on another thread, probably good to have "disabled" control on it.

packages/core/src/components/validation/validation.ts
josokinas: I don't mind either, you pick.

For the `state` prop - I agree this would be a breaking change.

Let's come back to that one when we later have to add any other "state", and then probably will need to do a breaking change...

packages/react/src/components/checkbox/checkbox.tsx
rabelloo: On the other hand several `Radio`s are not necessarily a `Fieldset`, usually that's used to group different inputs (with different `[name]`s).

Let me research `<fieldset>` usage a bit and think about this.

packages/react/src/components/form/form.stories.tsx
rabelloo: Here's one use case, a Stepper pattern where each Step is a Form.
Steps other than the current gets disabled.

image

You still want buttons to interact with the Form though.

rabelloo: You do have a point though, because `disabled` fields in a `Form` are simply ignored.

Which is why there's a `if (disabled) return;` in `Form`'s `onSubmit` which essentially prevents disabled `Form`s from being submitted even if the button isn't disabled.

Still, clicking the button should give you some form of feedback, rather than disabling the interaction altogether.

This is now stepping more into design territory though.

packages/core/src/components/validation/validation.ts
rabelloo: Cool, I'll go with `if` then.

I don't know we'll necessarily need a breaking change, presumably adding other `state`s like `success` would just be a `feat`.

But yea let's think about it then.
We could support both for a while and deprecate `state`.
Like
```ts
kind: 'error' | 'success';
@deprecated Use `kind` instead.
state: 'error';
```

And then just remove `state` on 2.0.0.

packages/react/src/components/form/form.stories.tsx
rabelloo: `Fieldset` already has `disabled`, `<fieldset>` supports it by default.

I don't need it because presumably a single `[name]` is just a single field, `<Radio>` is just an oddball.

Like I commented on the other thread, lt me research `<fieldset>` a bit more to see what makes more sense here.

packages/react/src/components/form/form.stories.tsx
josokinas: Yeah probably disabling a form should then be combined with (future) "loading" button state, that also is disabling clicking that button...

Design if your form though to me looks like a bit odd UX... I don't think "disabled" state of form elements is meant to be used like such, but what do I know... 😓

packages/react/src/components/index.ts
josokinas: So I guess what is actually needed also is for `.vscode/settings.json` to **disable** the autoformat of imports by the editor, right?

packages/core/src/components/validation/validation.ts
josokinas: Sure, should this be done _now_ then (in a separate PR)?

packages/react/src/components/form/form.stories.tsx
rabelloo: Oh I agree, it is weird. The step vanishing from view while you work on the next would be more usual.

Even with a loading state button, I still think we should not disable it and continue to emit events.

It's up to the consumer then to decide what to do with the `submit` events, in case it's loading.

They already control whether the button is loading or the form disabled after all.

packages/react/src/components/index.ts
josokinas: Oh I'm so sorry, so stupid...

I just now noticed that the change was from `export * from './field/field';` to `export * from './field';` 🤦‍♂️ Obviously it would then change position!

All good, nothing to worry about...

rabelloo: Yea it has also been reverted to `field/field` so should all be good.

packages/react/src/components/form/form.stories.tsx
rabelloo: So after some investigation I would agree that `button` inside `form` should be disabled, as should `input` and `select`.
That's what `fieldset` does after all.
Additionally `<a>` inside `fieldset` have `pointer-events: none;` set, so we should probably do the same for `form`.

If the customer wants an enabled button inside a disabled form they can either:
- manually set it `<Button disabled={false} />` 
- put the button outside of the form
- use a `fieldset` to disable other fields, but not button

Are you fine with the idea that `<Form disabled />` has the same behavior as `<fieldset disabled />`?
If so, I'll make it so `Button` also uses the `form` hook.

packages/react/src/components/checkbox/checkbox.tsx
rabelloo: After some research (comments on other threads), I think it's still nice to be able to disable the entire `Field` at will, instead of each `Radio` one by one.

It's a quite niche use case for `Radio`, as other inputs can be disabled directly just as easily.

Finally, it's better to have the lowest property determine the value, almost as if they cascade like CSS.

e.g.
```ts
<Form disabled>
  <Fieldset> ... </Fieldset>
  <Fieldset disabled={false}>
     <Field disabled>
       <Radio />
       <Radio />       
     </Field>
     <Field>
       <Textarea />
     </Field>
  </Fieldset>
</Form>
```

I don't expect to see such weird forms, just wanted to demonstrate the flexibility.

packages/react/src/components/form/form.stories.tsx
```diff
@@ -0,0 +1,122 @@
...
+          min={1}
+          max={31}
+        />
+        <Input type="number" name="month" placeholder="MM" min={1} max={12} />
+        <Input
+          type="number"
+          name="year"
+          placeholder="YYYY"
+          min={new Date().getUTCFullYear() - 150}
+          max={new Date().getUTCFullYear()}
```
josokinas, line 93: [Minor] Probably worth allowing minimum of 18, and a little more realistic maximum...

```suggestion
          min={new Date().getUTCFullYear() - 99}
          max={new Date().getUTCFullYear() - 18}
```

Altough I would very much prefer to live up to 150 years old!

packages/react/src/components/form/form.stories.tsx
josokinas: This looks good now, however why do you still need a parent Field?

packages/react/src/components/form/form.stories.tsx
```diff
@@ -0,0 +1,122 @@
...
+          Social media
+          <HelperText>Facebook, Twitter, TikTok, etc...</HelperText>
+        </Radio>
+        <Radio name="source">Search engine</Radio>
+        <Radio name="source">Word of mouth</Radio>
+        <Radio name="source">Other...</Radio>
+      </Fieldset>
+    </Field>
+
+    <Field>
```
josokinas, line 111: Probably this should also be a Fieldset?

packages/react/src/components/form/form.stories.tsx
```diff
@@ -0,0 +1,122 @@
...
+      <Checkbox name="agreement" required>
+        I agree to the <a>Privacy policy</a> and <a>Terms and conditions</a>.
```
josokinas, line 113: This is a little odd now, because on UI it does not do anything.

Should we rather not use any "links"?

```suggestion
        I agree to the Privacy policy and Terms and conditions.
```

packages/react/src/components/field/field.tsx
```diff
@@ -1,12 +1,61 @@
...
  */
-export const Field = ({ className, ...restProps }: FieldProps): JSX.Element => (
-  <div {...restProps} className={classy(c('field'), className)} />
-);
+export const Field = ({
+  className,
+  disabled,
+  onBlur,
+  onChange,
+  onInvalid,
```
josokinas, line 17: To align with all other components, keeping `className` below other props:

```suggestion
  disabled,
  onBlur,
  onChange,
  onInvalid,
  className,
```

packages/react/src/components/checkbox/checkbox.tsx
josokinas: I was originally thinking that a Field component is to wrap Input and Textarea components. It's like a helper component.

At least we state this in the component comment...

josokinas (approved):
Good job, this will be great if used in new apps when using Castor!

packages/react/src/components/form/form.stories.tsx
rabelloo: Hmm I was probably thinking about using `Validation` but that radio has no possible invalid state.
Will remove `Field` from there.

packages/react/src/components/field/field.tsx
```diff
@@ -1,12 +1,61 @@
...
+  ...restProps
+}: FieldProps): JSX.Element => {
+  const form = useForm();
+  const [field, setField] = useState<FieldState>(initial);
+
+  useEffect(() => update(form), [form]);
+
+  function initial(): FieldState {
```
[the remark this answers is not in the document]
rabelloo: It cannot, it needs to be hoisted to be used in `useState(initial)`.

I can change the others for consistency if you want, or we can make the hook slightly less efficient by creating another new function every render `useState(() => initial())`.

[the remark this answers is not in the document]
josokinas: Cool, fine to leave as it is.

packages/react/src/components/form/form.stories.tsx
rabelloo: How come? There's no need for a legend.

josokinas: OK, agree.

So probably you don't need to set anything, same as on this example: https://onfido.github.io/castor/?path=/docs/react-radio--playground#with-validation

Yeat again we do state Field is needed only for Input and Textarea components, isn't this a case anymore?

I guess if it does not work properly without using Field, then comments for Field component should also be improved.

packages/react/src/components/form/form.stories.tsx
rabelloo: Is this a service under-age people can't use? 😂 

The [oldest person to ever live](https://en.wikipedia.org/wiki/Oldest_people) died at 122 years of age.

I was trying to account for medicine improvements in the 21th century 🤣 

I mean, this is just a hypothetical form anyway, but if we're advocating for accessibility 🤷🏻

packages/react/src/components/form/form.stories.tsx
```diff
@@ -0,0 +1,121 @@
...
+      </Radio>
+      <Radio name="source">Search engine</Radio>
+      <Radio name="source">Word of mouth</Radio>
+      <Radio name="source">Other...</Radio>
+    </Fieldset>
+
+    <Field>
+      <Checkbox name="agreement" required>
+        I agree to the <a href="onfido.com">Privacy policy</a> and{' '}
+        <a href="onfido.com">Terms and conditions</a>.
```
josokinas, line 112: So seems like `<a>` does not work as expected?

It does not navigate to page, instead just toggles the checkbox.

packages/react/src/components/form/form.stories.tsx
rabelloo: Added links to Onfido's page, could also link to this repo (probably better).

packages/react/src/components/form/form.stories.tsx
rabelloo: Oh... Probably the missing protocol. I'll link to this repo instead and fix it.

packages/react/src/components/form/form.stories.tsx
rabelloo: It works without Field, but automatic Validation only works inside a Field.

I'll touch up the Field comments.

packages/react/src/components/form/form.stories.tsx
josokinas: No, still does not work...

rabelloo: Oh no, we have this in `.ods-input-indicator`: `pointer-events: none;`.

Why did we prevent events in there?

josokinas: It's this here: https://github.com/onfido/castor/pull/440

rabelloo: Ah it's stylistic. I'll change it to set the cursor instead then, and allow events.

josokinas: So it won't work.

Because you'd need to also keep a track of "disabled" state, that changes the cursor.

As described on PR, in such case when a link is to be added (altough not recommended), pointer events should be added back only on those links.

I believe I was looking at other design system examples, and pointer events were usually disabled.

packages/core/src/internal/indicator-container/indicator-container.scss
```diff
@@ -150,8 +150,8 @@
 
   // container children
   .ods-input-indicator ~ * {
+    cursor: pointer;
```
josokinas, line 153: This now breaks "disabled" state of both Checbox and Radio when label is set.

You would need some sort of "hack" like this:

```scss
.ods-input-indicator ~ * {
  pointer-events: none;

  a {
    pointer-events: auto;
  }
}
```

Which I personally don't like...

packages/react/src/components/form/form.stories.tsx
rabelloo: I see. Tricky.

packages/core/src/internal/indicator-container/indicator-container.scss
rabelloo: Right. Other alternatives are `z-index`, or

```scss
.ods-input-indicator ~ * {
  cursor: pointer;
}

.ods-checkbox:disabled,
.ods-radio:disabled {
    + .ods-input-indicator ~ * {
  cursor: not-allowed;
}
```

rabelloo: Why do you think the `a` rule is a hack though?

josokinas: Because I think it should be added to the `<a>` itself, as I mentioned - I think it's an edge case, and links by default should not be placed in labels.

Probably that your example could also work.

But please don't put `z-index` 😅

rabelloo: I don't have any favorites here, they all seem pretty meh.
Probably because links in checkbox labels are real meh.

The last option is as you said, require the `<a>` inside the label to allow pointer-events to itself again.
Would you prefer that? In the story itself?

packages/react/src/components/form/form.stories.tsx
rabelloo: Done, please review.

packages/core/src/internal/indicator-container/indicator-container.scss
josokinas: I'm not 100% sure also.

Maybe we could do that at least for now, and then come back to this if it does not work for users.

rabelloo: Sounds good. Updated with that.

packages/react/src/components/form/form.stories.tsx
josokinas: All seems good, thanks!

PR #456, opened 2021-03-22
fix: form disabled interactions

rabelloo:
## Purpose

Fixes an issue where form elements not wrapped by `<Field />` wouldn't represent the form disabled state.

## Approach

Better merge `form` and `field` state in `useField`, which is the only hook form elements use.

## Testing

Storybook.

## Risks

None.

### DIFF
--- a/packages/react/src/components/field/useField.ts
+++ b/packages/react/src/components/field/useField.ts
@@ -1,8 +1,18 @@
 import { FieldState } from '@onfido/castor';
 import { createContext, useContext } from 'react';
+import { useForm } from '../form/useForm';
 
 const FieldContext = createContext({} as FieldState);
 
-export const useField = () => useContext(FieldContext);
+export const useField = () => {
+  const form = useForm();
+  const field = useContext(FieldContext);
+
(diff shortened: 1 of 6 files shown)


packages/react/src/components/field/useField.ts
```diff
@@ -1,8 +1,18 @@
...
+  const form = useForm();
+  const field = useContext(FieldContext);
+
+  return { ...omitNullish(form), ...omitNullish(field) };
+};
 
 export const FieldProvider = FieldContext.Provider;
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const omitNullish = (obj: Record<string, any>) =>
```
josokinas, line 17: Odd naming, what's wrong with `omitNull`?

josokinas (approved):

packages/react/src/components/field/useField.ts
rabelloo: It also omits `undefined` 🤷🏻 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

PR #467, opened 2021-03-23
docs: display "color" selection options in description

josokinas:
## Purpose

"color" for Icon component has a list of available colors to apply, however Core story does not list colors properly.

## Approach

Manually set table properties for both Core and React stories so that lists of colors available are aligned.

## Testing

On Storybook, Icon component of each Core and React stories.

## Risks

N/A

### DIFF
--- a/packages/core/src/components/icon/icon.stories.ts
+++ b/packages/core/src/components/icon/icon.stories.ts
@@ -9,7 +9,12 @@ export default {
   title: 'Core/Icon',
   component: Icon,
   argTypes: {
-    color: { control: { type: 'select', options: colors } },
+    color: {
+      control: { type: 'select', options: colors },
+      table: {
+        type: { summary: colors.map((value) => `"${value}"`).join('|') },
+      },
+    },
   },
   args: {
     name: firstIconName,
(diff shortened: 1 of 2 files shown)


rabelloo (approved):

packages/core/src/components/icon/icon.stories.ts
```diff
@@ -9,7 +9,12 @@ export default {
   title: 'Core/Icon',
   component: Icon,
   argTypes: {
-    color: { control: { type: 'select', options: colors } },
+    color: {
+      control: { type: 'select', options: colors },
+      table: {
+        type: { summary: colors.map((value) => `"${value}"`).join('|') },
```
rabelloo, line 15: 💅🏻  We could set this function as `colors.toString` so that we can reuse it across files

rabelloo: Always commenting after merge 😬 

Don't worry about it

josokinas: Feel free to do a PR with this! 😏

PR #490, opened 2021-03-26
docs: add html field-label stories

josokinas:
## Purpose

## Approach

Adding stories to FieldLabel component.

Please note this excludes stories that include Validation component.

## Testing

On Storybook.

## Risks

N/A

### DIFF
--- a/docs/helpers/html.ts
+++ b/docs/helpers/html.ts
@@ -12,7 +12,7 @@ export const html = (
     `<${name}`,
     Object.entries(props).map(toAttribute).join('\n'),
     '>',
-    children,
+    Array.isArray(children) ? children.join('') : children,
     `</${name}>`,
   ]
     .filter(Boolean)
--- a/docs/helpers/html.ts
+++ b/docs/helpers/html.ts
@@ -12,7 +12,7 @@ export const html = (
     `<${name}`,
     Object.entries(props).map(toAttribute).join('\n'),
     '>',
-    Array.isArray(children) ? children.join('') : children,
+    children?.join?.('') || children,
     `</${name}>`,
   ]
     .filter(Boolean)
--- a/docs/helpers/html.ts
+++ b/docs/helpers/html.ts
@@ -12,7 +12,7 @@ export const html = (
     `<${name}`,
     Object.entries(props).map(toAttribute).join('\n'),
     '>',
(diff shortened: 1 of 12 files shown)


docs/helpers/html.ts
```diff
@@ -12,7 +12,7 @@ export const html = (
     `<${name}`,
     Object.entries(props).map(toAttribute).join('\n'),
     '>',
-    children,
+    Array.isArray(children) ? children.join('') : children,
```
josokinas, line 15: I'm not sure about this, what do you think @rabelloo maybe better just each time to join on each use?

rabelloo:
The snippet contains the output code rather than the Castor helpers, as it should tbf.

image

Not that this is a problem, but ideally it'd be `color('token-foo')` instead.

Any ideas?

packages/core/src/components/field-label/field-label.stories.tsx
```diff
@@ -0,0 +1,109 @@
...
+  Field({
+    children: [
+      FieldLabel({
+        ...props,
+        children: label,
+        for: id,
+      }),
+      Input({ id }),
+    ],
+  });
```
rabelloo, line 78: Remind me again why we do `input` outside of `label` with `[id]` and `[for]`?

rabelloo: i.e. 
image

rabelloo (approved):

packages/core/src/components/field-label/field-label.stories.tsx
josokinas: This was to demonstrate different ways of using field label with an input.

FieldLabel has examples of using `[for]`, where Input examples show as using nesting within `label`.

However, I have noticed that on HTML + CSS I used `[for]` everywhere... Will change.

josokinas:
> Any ideas?

I see the same output is for both storeis, it renders the output of a helper function.

TBH I'm not sure which is more correct, both are valid. And this might even be better, as it's an example for whether you use helper or you don't.

rabelloo:
> TBH I'm not sure which is more correct, both are valid. And this might even be better, as it's an example for whether you use helper or you don't.

That's a good point, I'm convinced.

PR #496, opened 2021-03-29
chore: use *.ts extension for core story files

josokinas:
## Purpose

Everything within "core" package should use `*.ts` but no `*.tsx` as it has no JSX.

## Approach

Use `*.ts` file extensions for core stories.

## Testing

Pipeline should pass, Storybook should function as normal.

## Risks

No restriction to continue adding `*.tsx` files to "core" package later, will need to look when reviewing PRs. However, minor.


rabelloo:
I tried doing this in the past, but it end up with stories not showing up in the left sidebar/nav.

I wonder if that was fixed in a Storybook patch or something 🤔

rabelloo (approved):

PR #533, opened 2021-04-06
docs: clean headings for better index

josokinas:
## Purpose

After GitHub introducing built-in index (table of content), our current headings included additional symbols resulting in odd appearance in UI:

Screenshot 2021-04-06 at 10 50 47

## Approach

Move "dot" alongside badge to a newline after the heading on root and each package.

## Testing

On https://github.com/onfido/castor/tree/docs/better-gh-index (root and also each package).

## Risks

N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,6 @@
-# Castor &middot; [![Storybook](https://raw.githubusercontent.com/storybooks/brand/master/badge/badge-storybook.svg)](https://onfido.github.io/castor/)
+# Castor
+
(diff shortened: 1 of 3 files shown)


rabelloo:
What if we just removed the dot? It doesn't look terrible imo

image

I quite like the Storybook link in the header 😢

josokinas:
I feel like it should be outside of the headline...

I will also be adding more badges (well, at least 1 more).

rabelloo:
I guess I was thinking of this style, which I find pleasing:

image


I'd love for us to do something like this when our logo is finalized:

image


Anyway, for now I guess what you suggest is alright.

rabelloo (approved):

PR #556, opened 2021-04-07
test: add visual regression support and spec all combinations

rabelloo:
## Purpose

Add support for visual regression tests.

## Approach

Use Cypress to navigate the Storybook instance and screenshot `All Combinations` stories.

## Testing

Locally, `yarn e2e` for headless or `yarn cy` for interactive.
In GitHub actions, there should be a new job inside the Test workflow.

## Risks

Requires a new set of tools (mainly Cypress).
Couldn't get `jest-image-snapshot` alone to work correctly directly, something about how Storybook build stories and the whole interaction of Jest + Storybook + Webpack + TypeScript + Babel.
Cypress is pretty good though, we could easily spin up any browser we want.


packages/react/src/components/button/button.tsx
```diff
@@ -10,7 +10,7 @@ export const Button: ButtonComponent = ({
 }: ButtonProps<'a'> | ButtonProps<'button'>) => {
   const { disabled } = useField();
 
-  const Element = 'href' in restProps ? 'a' : 'button';
+  const Element = (restProps as ButtonProps<'a'>).href ? 'a' : 'button';
```
rabelloo, line 13: Storybook passing `href: undefined` meant this rendered as `<a>`.

This is a fix not for our end customers, but for Storybook and ourselves.

.github/workflows/test.yml
```diff
@@ -21,3 +21,22 @@ jobs:
...
+        with:
+          node-version: 14
+
+      - run: |
+          yarn install --frozen-lockfile
+          yarn cypress cache path
+          yarn cypress cache list
+          yarn cypress verify
+          yarn cypress info
+          yarn e2e
```
rabelloo, line 42: Have to figure out a way to:

1. Export code coverage information to PR summary
2. Retain artifacts generated by Cypress (screenshots, video)
3. Cache `node_modules` for performance, could potentially be reused for all workflows and jobs
4. Whether to run a browser matrix here (with no mobile support) or use BrowserStack

rabelloo:
The workflow seems to be failing on an image diff, but hard to see why until artifacts are retained (screenshot).

I suspect it's a browser difference between mine locally (where baseline screenshots were taken) and the one on CI.

Will investigate later.

.github/workflows/test.yml
```diff
@@ -21,3 +21,22 @@ jobs:
 
       - name: Test packages
         run: yarn test
+
+  e2e-tests:
+    name: Run e2e tests
```
josokinas, line 26: [Minor] I think this is shortened as:

```suggestion
    name: Run E2E tests
```

.github/workflows/test.yml
josokinas: Can't answer any of these, would require researching same as for you.

However, because a preview of Surge allows to deploy multiple jobs on one PR, maybe it can be used to host both Cypress report and generated assets?

For example, on one of your apps we do store test assets as artifacts on GitLab (maybe [GitHub has similar approach](https://docs.github.com/en/actions/guides/storing-workflow-data-as-artifacts)?), but for the main branch it's stored on GitLab Pages.

Also, do not worry about caching `node_modules` for now. I agree this could be handy, however can be a separate thing.

.github/workflows/test.yml
```diff
@@ -21,3 +21,22 @@ jobs:
...
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions/setup-node@v2.1.4
+        with:
+          node-version: 14
+
+      - run: |
+          yarn install --frozen-lockfile
```
josokinas, line 37: Could we name steps, so they're better visible on UI (like [here](https://github.com/onfido/castor/runs/2321776409?check_suite_focus=true)):

```suggestion
      - name: Checkout Git repository
        uses: actions/checkout@v2

      - name: Setup Node
        uses: actions/setup-node@v2.1.4
        with:
          node-version: 14

      - name: Install Node dependencies
        run: yarn install --frozen-lockfile

      - name: Test packages
        run: |
```

.storybook/main.ts
```diff
@@ -33,6 +33,14 @@ module.exports = {
       '@onfido/castor-react': resolve(__dirname, '../packages/react/src'),
     };
 
+    // instrument source code if running e2e tests
```
josokinas, line 36: [Minor] If you agree with earlier, then:

```suggestion
    // instrument source code if running E2E tests
```

packages/react/src/components/button/button.tsx
josokinas: I just checked on UI and don't get this, how did you find this out?

e2e/visual-regression/all.e2e.ts
```diff
@@ -0,0 +1,23 @@
+const storyName = 'All Combinations';
```
josokinas, line 1: We also discussed testing all other states, because on some components it does not make sense to have `All Combinations`. Also, this would not include compositions.

Do you plan to still add to this PR, or it would be another one?

package.json
```diff
@@ -29,11 +29,20 @@
     "lint:style": "stylelint '**/*.{,s}css' --ignore-path .gitignore --max-warnings 0",
     "lint:prettier": "prettier . --check --ignore-path .gitignore",
     "test": "jest --coverage",
+    "cy": "concurrently -k yarn:e2e:serve yarn:cypress:open",
```
josokinas, line 32: Why did you need to use `concurrently` here?

package.json
```diff
@@ -29,11 +29,20 @@
...
+    "cypress:open": "cypress open",
+    "e2e": "concurrently -k --success first yarn:e2e:*",
+    "e2e:run": "wait-on http://localhost:6006 && cypress run -c video=true",
+    "e2e:serve": "NODE_ENV=e2e yarn start --ci --quiet",
+    "poste2e": "nyc report -r text -r text-summary",
+    "pree2e:run": "mkdir -p cypress",
```
josokinas, line 38: I personally like reading scripts as pre, then script itself, then post, so it's more clear the sequence.

What do you think:

```suggestion
    "e2e": "concurrently -k --success first yarn:e2e:*",
    "poste2e": "nyc report -r text -r text-summary",
    "pree2e:run": "mkdir -p cypress",
    "e2e:run": "wait-on http://localhost:6006 && cypress run -c video=true",
    "e2e:serve": "NODE_ENV=e2e yarn start --ci --quiet",
```

packages/react/src/components/button/button.tsx
rabelloo: I didn't, our automated visual regression tests did 🙊 

https://onfido.github.io/castor/iframe.html?id=react-button--all-combinations&viewMode=story

.github/workflows/test.yml
rabelloo: Yea we can use Cypress' action, seems pretty good.
- https://github.com/cypress-io/github-action
- https://www.cypress.io/blog/2019/11/20/drastically-simplify-your-testing-with-cypress-github-action

I'll try it out later.

.github/workflows/test.yml
rabelloo: I find it easy to understand `Run actions/setup-node` for example, but don't mind naming the steps.

rabelloo:
```suggestion
      - name: Checkout Git repository
        uses: actions/checkout@v2
        
      - name: Setup Node
        uses: actions/setup-node@v2.1.4
        with:
          node-version: 14
          
      - name: Install Node dependencies
        run: yarn install --frozen-lockfile
        
      - name: Run Cypress tests
        run: |
```

josokinas: Just to match with other actions.

Yet again, I don't mind then not using names, but it should be one PR removing naming from all actions rather than some actions having names, some not...

rabelloo: Renamed the last one slightly compared to your suggestion.

e2e/visual-regression/all.e2e.ts
rabelloo: I didn't plan for that, but it doesn't seem hard to introduce here.

But because the name isn't static, we either:
1. Screenshot all stories
2. Have an array of story names here
3. Write separate, specific spec/test file dedicated to each component

I feel like 3 is probably better, in which case it can be done after this PR. wdyt?

package.json
rabelloo: I want to serve the app and run cypress at the same time, in two separate but concurrent processes.

Cypress will wait until the specified url (the app) is ready to serve before running tests.

e2e/visual-regression/all.e2e.ts
josokinas: Sure, no problem, then title of this PR should probably be like: `test: add visual regression support for all combinations`.

package.json
rabelloo: Don't mind that at all. I never know when to use "pre/post" versus a simple `&&` either, I usually just avoid the script being too long.

packages/react/src/components/button/button.tsx
rabelloo: Same thing for me
image

e2e/visual-regression/all.e2e.ts
rabelloo: Well support is also there for other spec/tests, I just happened to check the support works for "All Combinations" stories.

Perhaps `test: add visual regression support and spec all combinations`?

packages/react/src/components/button/button.tsx
josokinas: I feel like this is some sort of Storybook bug, try this and scroll down: https://onfido.github.io/castor/?path=/docs/react-button--playground

e2e/visual-regression/all.e2e.ts
josokinas: Sure, thanks works 👍

packages/react/src/components/button/button.tsx
rabelloo: That works, so it seems that if the entrypoint story (the one on the url) is "All Combinations" it passes `{ href: undefined }`, otherwise it passes `{ }`.

Not sure if a bug, but definitely different behaviours depending on path.

rabelloo: Regardless, checking for value in the Button is probably better than checking for key, as we want this to render as `<button>`:

```tsx
<Button href={undefined} />
```

josokinas: Right, so anyways your change seems fine. Resolving this now.

josokinas, line 13: What do you think about having this change as a separate PR as a "react fix"?

Might be good to have such in a changelog?

package.json
```diff
@@ -29,11 +29,20 @@
     "lint:style": "stylelint '**/*.{,s}css' --ignore-path .gitignore --max-warnings 0",
     "lint:prettier": "prettier . --check --ignore-path .gitignore",
     "test": "jest --coverage",
+    "cy": "concurrently -k yarn:e2e:serve yarn:cypress:open",
+    "cypress:open": "cypress open",
+    "e2e": "concurrently -k --success first yarn:e2e:*",
+    "poste2e": "nyc report -r text -r text-summary",
+    "pree2e:run": "mkdir -p cypress",
```
josokinas, line 36: Just to be safe, probably better to also use `shx` here:

```suggestion
    "pree2e:run": "shx mkdir -p cypress",
```

package.json
josokinas, line 32: Not sure I like naming `cy` - maybe something like `cypress:interactive`? I know it's longer, but makes a little more sense when reading it.

josokinas (approved):
Looks good 👍

Mind also adding docs to `CONTRIBUTING.md` please?

packages/react/src/components/button/button.tsx
rabelloo: Hmm it shouldn't impact TypeScript users, but might JavaScript ones.

Yea I'll split it 👍🏻

rabelloo: https://github.com/onfido/castor/pull/576

package.json
rabelloo: `cypress:open`?

rabelloo: lol that's already the sub script. I mean `e2e:open`?

rabelloo: No that doesn't work either, it'll be confusing for `e2e` which can no longer run `npm:e2e:*`

rabelloo: Perhaps simply `cypress`?

josokinas: Just to be clear, you want this as a script only to avoid having 2 terminal window open right?

But when you think about it, mostly if you run tests you'd also developing so you _would_ have Storybook running? So can it just be instruction to run `yarn cypress open` whiles running a Storybook instead of script(s)?

rabelloo: Correct, single terminal, two processes.

I guess it could, this is usually a good script for test engineers who rarely open Storybook alone.

I don't mind removing `cy` and `cypress:open` if you think they're unnecessary.

rabelloo:
> Mind also adding docs to CONTRIBUTING.md please?

Please do let me know how it looks.

CONTRIBUTING.md
```diff
@@ -56,6 +56,50 @@ You can run a Storybook instance locally with:
...
+
+You can also filter to a single file if you want, like:
+
+    yarn jest --watch my-file
+
+#### UI tests (including visual-regression)
+
+Those are found in `./e2e`, in its root or inside folders, but mandatorily with a `.e2e.ts` file suffix.
+
+They are written in Cypress.
```
josokinas, line 81: [Minor]

```suggestion
They are written in [Cypress](https://www.cypress.io/).
```

package.json
josokinas: I think we could start with having command instruction on docs (the ones you just added), and see if we need to have scripts?

So we would have more time for the most difficult thing - naming :trollface:

rabelloo: Sounds good.

rabelloo: Oh wait, it needs to build Storybook with an environment variable, `NODE_ENV=e2e`.

Do we want the instruction to read like:

```
Run Storybook in e2e mode:
   NODE_ENV=e2e yarn start

Open Cypress:
    yarn cypress open
```

josokinas: So I suggest this:

- remove scripts `cy` and `cypress:open`
- instruct for such to serve the app with `yarn e2e:serve` and then run `yarn cypress open`

How does this sound?

rabelloo: Done.

CONTRIBUTING.md
```diff
@@ -56,6 +56,60 @@ You can run a Storybook instance locally with:
...
+
+   yarn e2e:serve
+
+2. Then in another terminal open Cypress with:
+
+   yarn cypress open
+
+Alternatively you could use a single terminal with:
+
+    yarn concurrently -k -n ,cypress:open yarn:e2e:serve "yarn cypress open"
```
josokinas, line 103: Is this command correct, do you need a `,` here?

josokinas (approved):

rabelloo:
Alright this is now ready 🎉 

Environment differences (between local and CI for example) are enough for images not to match.
So I have added another job `screenshot-baselines` to provide us with those when they mismatch.

Artifacts are kept nicely. If `e2e` fails, we can download those images and update the baselines.

josokinas:
That's great!

One thing to note though, on a core screenshot is not of rendered UI but of code snippets instead, so essentially of this: https://onfido-castor-preview-pr-556.surge.sh/iframe.html?id=core-button--all-combinations&args=&viewMode=story

What do you reckon we should do, probably remove tests for core at least for now?

CONTRIBUTING.md
rabelloo: Yea it looks quite weird, we need to `-n` name the second command, but not the first.

```ts
concurrently({
  k: true, // kill other processes when the first one dies
  n: [, 'cypress:open'], // nulls are ignored
  commands: [
    'yarn:e2e:serve', // auto named 'e2e:serve'
    'yarn cypress open' // can't auto name
  ]
});
```

So when they run they're logged nicely like

```
[e2e:serve] Running on http://localhost:6006
[cypress:open] Starting Cypress...
```

rabelloo:
Yea I noticed that, our decorator that transforms that string into a DOM node doesn't run in that scenario :/

I think keeping it is fine, we're essentially snapshoting (like, comparing markup) for now.

Making `core` stories work as iframes is something we want either way, and it's not too hard to fix I reckon.

josokinas:
OK, fine, good to go then 🚀

rabelloo:
Next step is to explore using CI matrixes for other browsers, or a 3rd party tool like BrowserStack.

We'll probably need baseline screenshots for each browser though, as they have subtle rendering differences.

PR #587, opened 2021-04-16
fix(core): search includes icon and input if not already added

rabelloo:
## Purpose

Allow `@include castor.Search` to function without necessarily adding Icon or Input too.

## Approach

By using `includes.the()` we make sure it's not included twice.

## Testing

Open up any example, leave only castor.Search styles and render a Search component.

## Risks

`includes` management seems to be getting trickier.

There's another possible approach which would require a refactor of other usages:

```
@mixin Search() {
  @include icon.Icon();
  @include input.Input();

  @if not includes.the('search') {
    // search styles ...
  }
}

@mixin Input {
  @if not includes.the('input') {
    // input styles ...
  }
}
```

Might be easier to reason, happy to do that in a later PR.


josokinas:
I agree this is not sustainable.

I think this change makes things even more confusing.

What if the fix would be the following:

- `includes` moved to`mixins`, directly is used for e.g. internals, animations, etc.
- we have a new mixin `component`, that is used for every single component creation (similar to how themes are created), which then always uses `includes` to indicate if component has been already included
- `includes` is no longer used in `components` mixin (no longer needed)

Then last part would be how to define what includes a component needs. This could either be manually entered on a component, or defining when using the new `component` mixin.

What do you think?

rabelloo:
Explored that a little bit, the issue is the nesting depth lint
image

rabelloo:
Also, applying mixins from a list of strings reads way more poorly than what we do for themes which is just a class name

rabelloo:
Pushed a refactor suggestion, take a look and let me know how you feel

packages/core/src/components/field-label/field-label.scss
```diff
@@ -1,6 +1,13 @@
 @use '../../helpers';
+@use '../../mixins';
 
 @mixin FieldLabel() {
+  @if not mixins.includes('FieldLabel') {
+    @include _fieldLabel();
+  }
+}
+
+@mixin _fieldLabel() {
```
josokinas, line 10: Naming becomes odd now, maybe each mixin should be identical to actual mixin but prefixed as internal, so here it would be `_FieldLabel` (and on all other components).

packages/core/src/components/search/search.scss
```diff
@@ -1,6 +1,18 @@
 @use '../../helpers';
+@use '../icon/icon';
+@use '../../mixins';
```
josokinas, line 3: I think sorting is wrong, `../../` goes before `../`.

```suggestion
@use '../../mixins';
@use '../icon/icon';
```

josokinas (approved):
I prefer this refactor rather than the previous fix, thanks!

packages/core/src/components/field-label/field-label.scss
rabelloo: Yea I don't think there's a good way to name it :/

I like the same name idea though, will rename everything.

packages/core/src/internal/indicator-container/indicator-container.scss
```diff
@@ -2,6 +2,12 @@
 @use '../../mixins';
 
 @mixin IndicatorContainer() {
+  @if not mixins.includes('internal: IndicatorContainer') {
+    @include _indicatorContainer();
+  }
+}
+
+@mixin _indicatorContainer() {
```
josokinas, line 10:
```suggestion
    @include _IndicatorContainer();
  }
}

@mixin _IndicatorContainer() {
```

rabelloo: Missed one, thanks 🙈

PR #589, opened 2021-04-20
feat: generate themes with theo

josokinas:
## Purpose

In order to (later) generate themes not only for web, but also for iOS and Android, we need one source of aliases & themes in a format that can be used to generate multi-platform variants.

## Approach

Using [Theo](https://github.com/salesforce-ux/theo) to generate (for now) web themes, following the https://github.com/onfido/castor/pull/136 development.

## Testing

Tested locally, pipeline should also pass, preview generated.

## Risks

N/A


.vscode/settings.json
```diff
@@ -1,11 +1,32 @@
 {
   "[handlebars]": {
-    "editor.tabSize": 2
+    "editor.formatOnSave": false
```
josokinas, line 3: Prettier is not a good friend with Handlebars, so I am disabling the formatting on a save for this type. Also, I do enable auto formatting explicitly on file types, just to have more control over it.

package.json
```diff
@@ -15,6 +15,8 @@
     "start": "start-storybook -p 6006",
     "generate": "concurrently yarn:generate:*",
     "generate:tokens": "esbuild scripts/generate-tokens.ts --bundle --platform=node --target=es2019 --format=cjs --external:theo | node",
+    "generate:theme:day": "esbuild scripts/generate-theme.ts --bundle --platform=node --target=es2019 --format=cjs --external:theo | node",
+    "generate:theme:night": "esbuild scripts/generate-theme.ts --bundle --platform=node --target=es2019 --format=cjs --external:theo | node",
```
josokinas, line 19: I do get the theme name from script name, e.g. "day" or "night".

This however then requires having two scripts set, even though their executions are the same.

I could have done only with one script where I keep an array of theme names, however I feel like this is more powerfull e.g. you could run a script for only one theme.

packages/core/src/theme/tokens.json
```diff
@@ -1,80 +1,12 @@
 {
-  "aliases": {
```
josokinas, line 2: I had to separate this to a indipendent file, because otherwise theme importing `tokens.json` also processes the token props.

packages/core/src/theme/tokens.json
```diff
@@ -1,80 +1,12 @@
...
-    "color-error-muted-300": "239, 169, 169"
-  },
+  "imports": ["./aliases.json"],
   "props": [
+    {
+      "type": "duration",
+      "category": "transition",
+      "name": "duration",
+      "value": "0.15s"
+    },
```
josokinas, line 9: I've decided to move this to tokens instead of each theme, as we most likely won't be having separate values.

Also, at any point you can just override this with a custom theme.

.vscode/settings.json
rabelloo: I usually keep these locally as user preferences, but I don't mind them on the repo, consistency.

package.json
rabelloo: Yea I like that, more flexible.

packages/core/src/theme/tokens.json
rabelloo: That's fine, but the naming is confusing.
I see tokens as the actual value pointers/references, e.g. `token: 2px`
Then aliases reference that token, e.g. `alias: var(--token)`

Are you fine with swapping the name on these files, or do you have another suggestion?

scripts/generate-theme.ts
```diff
@@ -0,0 +1,56 @@
...
+} from 'theo';
+
+const npmLifecycleEvent = process.env.npm_lifecycle_event;
+
+if (!npmLifecycleEvent)
+  throw new Error('Unable to retrieve npm lifecycle event');
+
+const theme = npmLifecycleEvent.split(':').pop();
+
+const rootPath = resolve('.');
```
rabelloo, line 18: Isn't this the same as `_dirname`, but flakier?

packages/core/src/theme/tokens.json
josokinas: I'd like to keep the Theo naming...

Let's forget now the CSS tokens, this will also be used for iOS & Android, so e.g. even Android has ability to alias but iOS does not... It the outcome will be different on each platform.

.vscode/settings.json
josokinas: When I got onto a new machine I found it odd my usual setup wasn't working anymore... Obviously I forgot to move the config. But then I think having it on a repo makes more sense, everyone contributing would use the same...

scripts/generate-theme.ts
```diff
@@ -0,0 +1,56 @@
...
+  transform: {
+    type: 'raw',
+    file: srcPath,
+  },
+  format: {
+    type: 'scss',
+  },
+})
+  .then((data) =>
+    writeFileSync(destPath, data.replace(/\$\$THEME/g, `${theme}`))
```
rabelloo, line 45: `$$THEME` is an interesting choice, usually handlebars interpolations are like `{{THEME}}`.

I don't mind either, just curious 🤔

rabelloo (approved):

scripts/generate-theme.ts
josokinas: So I can't use `{{THEME}}` as it gets interpolated during the Theo run.

I do need to replace this after Theo does it's job.

I don't mind having a different naming, `$$THEME`  was just some random choice...

packages/core/src/theme/tokens.json
rabelloo: I see. Designers also refer to them as "design tokens", adding to the confusion.
Let's not use the word "alias" anymore then, in chat and verbally, we can call them "base/palette tokens" and "theme tokens".

scripts/generate-theme.ts
rabelloo: A-ha! I thought it was Theo doing the interpolation, and this was some sort of instruction. Makes more sense now.

I'm fine with `$$`. Although, can we not have Theo interpolate the theme? Would it require adding a new transform?

packages/core/src/theme/tokens.json
josokinas: I would say these are Theo aliases.

Then we have base tokens, then theme tokens.

During generation of base tokens Theo aliases are used.

During generation of theme tokens also Theo aliases are used.

However from design perspective theme tokens alias to base tokens.

So I see this as separate terminology for design and engineering sides.

.vscode/settings.json
rabelloo: Assuming they use VS Code, but yea. I like them on the repo, it's cool.

scripts/generate-theme.ts
josokinas: I've replacing now before passing format source to Theo.

PR #591, opened 2021-04-20
feat(core): themes set color-scheme

rabelloo:
## Purpose

Currently the browser appearance does not adapt to Castor's set theme.

Most visible on a scrollbar:
image

Make it so ([browsers that support it](https://developer.mozilla.org/en-US/docs/Web/CSS/color-scheme#browser_compatibility)) do:
image

## Approach

Extend the theme mixin to respond to a `dark` type, adding `color-scheme` accordingly.

## Testing

Look at the output `themes/night.css` and `day.css`.

## Risks

Some browsers like Firefox don't support it, but it should simply be ignored.

### DIFF
--- a/packages/core/src/theme/theme.scss
+++ b/packages/core/src/theme/theme.scss
@@ -2,14 +2,17 @@
 @use '../mixins';
 @use './tokens' as castor;
 
-// ...$types: 'class' | 'raw';
+// ...$types: 'class' | 'dark' | 'raw';
 @mixin theme($name, $types...) {
   $class: list.index($types, 'class');
+  $dark: list.index($types, 'dark');
   $raw: list.index($types, 'raw');
 
   $selector: if($class, '.castor-theme--#{$name}', null);
 
   @include mixins.wrap-if($selector) {
+    color-scheme: if($dark, 'dark', 'light');
+
     @if not $raw {
       @include castor.tokens();
     }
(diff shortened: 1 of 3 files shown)


josokinas:
I am not to keen to have such on a theme generator/mixin, probably this is something that is to be added on a theme only?

That is, we would add on "day", then on "night" theme individually.

Otherwise when creatinga custom theme you get this out of the box, but would you expect something to be set?

Also, I don't understand what this mean "only dark keyword" and "only light keyword" that are not supported?

Screenshot 2021-04-20 at 16 26 39

rabelloo:
I would expect other themes to set it too, otherwise they might get "light" scrollbars on a "dark" theme and be confused.

I think that means `color-scheme: only dark` to state explicitly that `light` is not supported.
Seems like an unusual feature though.

josokinas:
Cool, but that would mean that when creating a custom theme you'd need to decide whether it is "light" or "dark". Can it be neutral, not setting anything?

I still think that setting this on our themes is OK, but not including in a theme mixin...

josokinas:
I would lean towards having in `day.scss` and `night.scss`, these would also have to be adopted to Theo generation after https://github.com/onfido/castor/pull/589.

But I also don't mind if a theme mixin accepts `'dark' | 'light'` and sets or not if nothing provided.

For such we would then however need to "Creating Theme" docs on Storybook.

rabelloo:
I'll go with the simpler option.

josokinas (approved):
TBH I wish our themes would also be called "light" and "dark"...

PR #592, opened 2021-04-20
fix(core): monospace font

rabelloo:
## Purpose

Currently the monospace font definition is wrapped in quotes, so the font-family does not apply.

image

image

This unwraps the string so that it works.

## Approach

`string.unquote` in Sass.

## Testing

`@include castor.font('400-monospace')` should apply correctly.

## Risks

None.

### DIFF
--- a/packages/core/src/helpers/font/font.scss
+++ b/packages/core/src/helpers/font/font.scss
@@ -81,7 +81,7 @@ $_types: (
     text-transform: uppercase,
   ),
   'mono': (
-    font-family: '"Roboto Mono", monospace',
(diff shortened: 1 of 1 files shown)


josokinas:
I don't seem to understand the issue here...

This PR:

Screenshot 2021-04-21 at 10 15 39

Current `main` branch:

Screenshot 2021-04-21 at 10 16 04

This seems to be working on Storybook at least?

josokinas (approved):
OK, no worries, I now realised we're using JS helper on Storybook... 🙈

rabelloo:
We might need a better way to test the Sass helpers, we don't use monospace in the examples so we didn't catch this one.

PR #607, opened 2021-04-23
chore: configure rulers for vscode

josokinas:
## Purpose

We do try to fit code into 80 characters long (which is done by linters), however having a visual line would also allow to keep such for other stuff like inline comments.

## Approach

Add ruler of 80 characters to VS Code.

## Testing

Locally on VS Code.

## Risks

Only affects users that would use VS Code.


rabelloo (approved):
This change is a little bit on the opinionated side, unlike the others.
I'm personally fine with the ruler, just something to keep in mind that it might annoy other contributors.

PR #614, opened 2021-04-23
docs(react): refactor search component stories

josokinas:
## Purpose

Refactor React stories for Search component.

## Approach

Use `reactMatrix`, remove `storyOf`.

## Testing

On Storybook, React stories for Search component.

## Risks

@rabelloo to confirm if something else needs to be removed.


rabelloo (approved):
I think that's all of it, if we missed a file we can delete it later.

PR #615, opened 2021-04-23
fix(react): form onSubmit values

rabelloo:
## Purpose

The current approach will read `id` off elements if they have one, instead of reading their `names`.

That's because of how `<form>.elements` work.

image

## Approach

Instead of using the `elements` key, just list them as an array and filter the ones that have the `name` property set.
It's actually much more efficient this way, and we can remove some code.

## Testing

Updated unit tests.

## Risks

None.

### DIFF
--- a/packages/react/src/components/form/getFormValues.ts
+++ b/packages/react/src/components/form/getFormValues.ts
@@ -13,14 +13,10 @@
  */
 export const getFormValues = <T extends Result>(form: HTMLFormElement): T =>
   Object.fromEntries(
-    Object.entries(form.elements).filter(hasName).map(elementValue)
-  );
+    (Array.from(form.elements) as HTMLInputElement[])
+      .filter((el) => !!el.name)
+      .map((el) => [el.name, el.value])
+  ) as T;
 
-const hasName = ([key]: Entry) => !isDigit.test(key);
-const elementValue = ([key, element]: Entry) => [key, valueOf(element)];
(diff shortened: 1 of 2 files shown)


packages/react/src/components/form/getFormValues.ts
```diff
@@ -13,14 +13,10 @@
  */
 export const getFormValues = <T extends Result>(form: HTMLFormElement): T =>
   Object.fromEntries(
-    Object.entries(form.elements).filter(hasName).map(elementValue)
-  );
+    (Array.from(form.elements) as HTMLInputElement[])
```
josokinas, line 16: If you set this as `HTMLInputElement`, does that work with other form elements?

josokinas (approved):

packages/react/src/components/form/getFormValues.ts
rabelloo: It does, all we need is elements with `.value`, all all `form.elements` should have them.
It's not the most precise or correct interface/type to use, but it's not terribly wrong either.
I suppose we could do `as { value: string }` 🤷🏻

josokinas: Maybe it woul be better?

IDK, just seems that you only expect `<input>` here...

rabelloo: Fair enough, created a new interface.

PR #619, opened 2021-04-27
feat(core): improve font monospace list

rabelloo:
## Purpose

When "Roboto Mono" is not available, the default `monospace` font in Chrome/MacOS is [Courier](https://en.wikipedia.org/wiki/Courier_(typeface)), which is a serif font and looks nothing like Roboto.

So, this improves the monospace list in case "Roboto Mono" isn't available.

## Approach

Add `Consolas` for Windows and `Menlo` for MacOS.

## Testing

Use monospace font helpers but don't provide "Roboto Mono".

## Risks

None.

### DIFF
--- a/packages/core/src/helpers/font/font.scss
+++ b/packages/core/src/helpers/font/font.scss
@@ -81,7 +81,7 @@ $_types: (
     text-transform: uppercase,
   ),
   'mono': (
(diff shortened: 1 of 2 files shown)


josokinas (approved):
CC @Subcide

PR #622, opened 2021-04-28
feat(react): deprecate "children" for adding input/textarea label

josokinas:
## Purpose

Using "children" prop to display a label on React's Input and Textarea is not very accurate, because "children" on Textarea has another meaning, also e.g. for Select component this also will not work - options will need to be provided via children instead.

## Approach

Deprecate usage of "children" prop on Input and Textarea components, suggest component composition instead.

## Testing

N/A

## Risks

N/A

### DIFF
--- a/packages/react/src/components/form/form.stories.tsx
+++ b/packages/react/src/components/form/form.stories.tsx
@@ -35,27 +35,46 @@ interface Values {
 export const Playground: Story<FormProps<Values>> = (props) => (
   <Form {...props}>
     <Field>
-      <Input name="firstName" required>
-        First name <Asterisk aria-label="required" />
-      </Input>
+      <Input
+        name="firstName"
+        label={
+          <>
+            First name <Asterisk aria-label="required" />
+          </>
+        }
+        required
+      />
       <Validation state="error" if="valueMissing">
         Please fill in this field
       </Validation>
     </Field>
 
     <Field>
-      <Input name="lastName" required>
-        Last name <Asterisk aria-label="required" />
-      </Input>
+      <Input
+        name="lastName"
+        label={
+          <>
+            Last name <Asterisk aria-label="required" />
+          </>
+        }
+        required
+      />
       <Validation state="error" if="valueMissing">
         Please fill in this field
       </Validation>
     </Field>
 
     <Field>
-      <Input type="email" name="email" required>
-        Email address <Asterisk aria-label="required" />
-      </Input>
+      <Input
+        type="email"
+        name="email"
+        label={
+          <>
+            Email address <Asterisk aria-label="required" />
+          </>
+        }
+        required
+      />
       <Validation state="error" if="valueMissing">
         Please fill in this field
       </Validation>
--- a/packages/react/src/components/form/form.stories.tsx
+++ b/packages/react/src/components/form/form.stories.tsx
@@ -35,46 +35,36 @@ interface Values {
 export const Playground: Story<FormProps<Values>> = (props) => (
   <Form {...props}>
     <Field>
-      <Input
-        name="firstName"
-        label={
-          <>
-            First name <Asterisk aria-label="required" />
-          </>
-        }
-        required
-      />
+      <FieldLabel>
+        <span>
+          First name <Asterisk aria-label="required" />
+        </span>
+        <Input name="firstName" required />
+      </FieldLabel>
       <Validation state="error" if="valueMissing">
         Please fill in this field
       </Validation>
     </Field>
 
     <Field>
-      <Input
-        name="lastName"
-        label={
-          <>
-            Last name <Asterisk aria-label="required" />
-          </>
-        }
-        required
-      />
+      <FieldLabel>
+        <span>
+          Last name <Asterisk aria-label="required" />
+        </span>
+        <Input name="lastName" required />
+      </FieldLabel>
       <Validation state="error" if="valueMissing">
         Please fill in this field
       </Validation>
     </Field>
 
     <Field>
-      <Input
-        type="email"
-        name="email"
-        label={
-          <>
-            Email address <Asterisk aria-label="required" />
-          </>
-        }
-        required
-      />
+      <FieldLabel>
+        <span>
+          Email address <Asterisk aria-label="required" />
+        </span>
+        <Input type="email" name="email" required />
+      </FieldLabel>
       <Validation state="error" if="valueMissing">
         Please fill in this field
       </Validation>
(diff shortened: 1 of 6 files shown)


packages/react/src/components/form/form.stories.tsx
```diff
@@ -35,27 +35,46 @@ interface Values {
...
-      </Input>
+      <Input
+        name="firstName"
+        label={
+          <>
+            First name <Asterisk aria-label="required" />
+          </>
+        }
+        required
+      />
```
rabelloo, line 46: This is what I had in mind when we discussed fragments being used in the label.
Personally, I think composition looks much cleaner:

```tsx
<Label>
  First name <Asterisk aria-label="required" />
  <Input name="firstName" required />
</Label>
```

I wonder again if we should even have `label` at all.
The only use case that makes sense to me is:

```tsx
<Input name="firstName" label="First name" required />
```

And even that is a stretch. Doesn't seem much easier or cleaner than:

```tsx
<Label>
  First name
  <Input name="firstName" required />
</Label>
```

packages/react/src/components/input/input.tsx
```diff
@@ -44,4 +46,12 @@ export const Input = withRef(
 );
 Input.displayName = 'Input';
 
-export type InputProps = BaseProps & JSX.IntrinsicElements['input'];
+export type InputProps = BaseProps &
+  Omit<InputElementProps, 'children'> & {
+    /** @deprecated Use `label` prop instead */
+    children?: InputElementProps['children'];
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    label?: any;
```
rabelloo, line 54:
```suggestion
    label?: React.ReactNode;
```

packages/react/src/internal/field-label-wrapper/field-label-wrapper.tsx
```diff
@@ -3,17 +3,17 @@ import React from 'react';
...
-  children: { element, children },
+  children: { element, label },
 }: FieldLabelWrapperProps): JSX.Element => {
-  if (!children) return element;
+  if (!label) return element;
 
   return (
     <FieldLabel htmlFor={id}>
-      <span>{children}</span>
+      <span>{label}</span>
```
rabelloo, line 16: Remind me again why we need the `<span>`? 🤔

packages/react/src/components/input/input.tsx
josokinas: Hmm, you think so?

I was looking at what usually `children` is, which essentially was what used previously, and seems like it's just `any`?

packages/react/src/internal/field-label-wrapper/field-label-wrapper.tsx
josokinas: It's because of the [flow in component](https://github.com/onfido/castor/blob/feat/input-textarea-label/packages/core/src/components/field-label/field-label.scss#L16), we expect usually a single label (e.g. just a string), and optional element.

For such cases where label is of multiple elements, we need to wrap in `<span>` to make it one element.

packages/react/src/components/form/form.stories.tsx
josokinas: Similar to previous reply, it would need to still be wrapped in `<span>` (where a "required" indicator is added).

My intentions here was actually to demonstrate yet another way of setting a label. So previously it was with only a string, then with component components, and this - with using "label" prop but as JSX.

packages/react/src/components/input/input.tsx
rabelloo: Yea pretty sure
```ts
interface DOMAttributes<T> {
        children?: ReactNode;
```

rabelloo: There are other instances around other files, I'd recommend importing `{ ReactNode } from 'react'` and updating them all.

josokinas: Yep, done that locally.

packages/react/src/components/form/form.stories.tsx
rabelloo: Right

```tsx
<Label>
  <span>First name <Asterisk aria-label="required" /></span>
  <Input name="firstName" required />
</Label>
```

The `label` prop still looks a bit magic, like, you give it a Fragment but it renders in a `span`.
Wouldn't it make more sense to actually require the wrapping `span` explicitly?

```tsx
<Input
  name="firstName"
  label={
    <span>
      First name <Asterisk aria-label="required" />
    </span>
  }
  required
/>
```

In which case isn't composition still cleaner?

I'm not trying to shoot `[label]` down, I'm really trying to understand where it fits better than `<Label>`.

josokinas: You're right, now when it's just a prop it could just render whatever passed... Totally makes sense.

Main worry - it's a breaking change.

What if we leave this for now, and I'll put a reminder to remove when making a breaking change (that is, when removing "children" prop support).

rabelloo: Can't we do something like

```tsx
const Input = ({ children, label }) =>
  <Wrapper>
    {{
      label: children ? <span>{children}</span> : label,
    }}
  </Wrapper>
```

And one final push, is having `[label]` even worth it?
Isn't composition always just better?
Or do you feel like providing several options is what makes it better?

josokinas: You win, refactored that part, so now only deprecating using "children".

rabelloo: I didn't want to win 😅  I mean not necessarily.
I'd be happy to agree that flexibility is an advantage in and of itself.

If you're happy without `[label]` then cool.
If you're compromising, then feel free to push back.

packages/react/src/components/input/input.tsx
```diff
@@ -48,9 +46,8 @@ Input.displayName = 'Input';
 
 export type InputProps = BaseProps &
   Omit<InputElementProps, 'children'> & {
-    /** @deprecated Use `label` prop instead */
+    /** @deprecated Use component composition instead */
```
rabelloo, line 49: Can we add an example here, or a link to storybook?
Not everyone might know what we mean by "component composition".

rabelloo (approved):

packages/react/src/components/input/input.tsx
josokinas: Added examples, altought not their nor deprecation message is shown on editor, it just says "deprecated"...

rabelloo: That's fine, it'll be hard to see the deprecation notice in the IDE regardless, because of how JSX children are shown.

We'll have to do a good job at notifying this change.

PR #635, opened 2021-05-03
fix(react): form values onChange and onSubmit

rabelloo:
## Purpose

Fix `Form` to read values properly depending on element attributes.

## Approach

Check for element attributes and resolve value according to each of them.

## Testing

Completely revamped the unit test suite.
Manual testing can be done on the story.

## Risks

None.


josokinas (approved):
This is awesome 👏

PR #662, opened 2021-05-10
build: use tokens (incl. themes) from separate package

josokinas:
## Purpose

Use new `@onfido/castor-tokens` package.

## Approach

Upon `yarn install` integrate Castor Tokens.

This is because we cannot refer to `node_modules` easily on Sass.

## Testing

CI should succeed, Storybook should work as expected.

## Risks

N/A


packages/core/src/helpers/border-radius/border-radius.scss
```diff
@@ -1,10 +1,9 @@
+@use '@onfido/castor-tokens' as tokens;
```
josokinas, line 1: This is an odd one.

Basically, running with `sass-cli` does not find `node_modules` imports if using:

```scss
@use '~@onfido/castor-tokens';
```

Looks like a solution is to use [load paths](https://sass-lang.com/documentation/at-rules/use#load-paths), however then it should not have `~`.

But then using a helper in app fails, e.g. here: https://github.com/onfido/castor/pull/662/checks?check_run_id=2622081236

Not sure yet what to do here TBH.

josokinas:
@rabelloo this is ready for initial review, however I have an issue currently that is breaking so will need to still sort that one out...

packages/core/src/helpers/border-radius/border-radius.scss
rabelloo: afaik that's Parcel 1 that requires the starting `~`, Parcel 2 does not.
Don't remember what Webpack does.

This will be tricky 😬

rabelloo: From our docs, Webpack seems to require the tilde `~` too 🙈

packages/core/src/index.scss
```diff
@@ -1,3 +1,3 @@
+@forward '~@onfido/castor-tokens'; // TODO: remove after Castor and Castor Tokens split BREAKING CHANGE
```
rabelloo, line 1: Is the goal that people import tokens and Color types directly from `castor-tokens` instead?

I don't mind leaving this here permanently, as a re-export.

rabelloo (approved):

packages/core/src/helpers/border-radius/border-radius.scss
josokinas: I've refactored all that part, now I'm integrated the whole `src` upon `yarn install`. So Castor Tokens becomes part of Core source.

packages/core/src/index.scss
josokinas: I would expect this, yes. It would be a breaking change. But needs more thinking here..

docs/colors/tokens.scss
```diff
@@ -1,6 +1,6 @@
-@use '~@onfido/castor';
+@use '~@onfido/castor' as theme;
 
 :export {
-  @include castor.day();
-  @include castor.night();
+  @include theme.day();
+  @include theme.night();
```
rabelloo, line 5: 🤓  [no action required]
Consumers would probably use `@include castor.day` so I don't think it's a big deal to have this import not renamed here.
Just thinking out loud, I don't mind either version.

packages/core/src/helpers/border-radius/border-radius.scss
```diff
@@ -1,10 +1,9 @@
 @use 'sass:list';
-
-$_allowed-sizes: ('small', 'medium', 'large', 'full');
+@use '../../integrated/tokens' as tokens;
```
rabelloo, line 2: 🤔  Would we be able to do `@use '@onfido/castor-tokens';` if we copied the files into `packages/tokens` instead?
That would make it part of the workspace, and we can gitignore the entire folder.
We might only need a `package.json` to name it.

packages/core/src/index.scss
rabelloo: Alright we'll cross that bridge when we get there.
I personally prefer importing all of it from `@onfido/castor`.

packages/core/src/helpers/border-radius/border-radius.scss
josokinas: No, it's the same. On Sass you can't, we cannot predict what setup users have on their side to handle npm packages...

rabelloo: Ah yea forgot about the tilde thing, nvm

PR #698, opened 2021-05-21
refactor: import from package rather than relative path

josokinas:
## Purpose

Keep imports consistent.

## Approach

If import is exported via the package, import directly from it. Otherwise use relative path.

Only imports via relative paths are:
- internal components
- doc stuff
- in case it's a sibling file (e.g. if component has its own utilities)
- React utilities (not exported)

## Testing

Pipeline should be green, Storybook should function as expected.

## Risks

N/A

### DIFF
--- a/packages/core/src/utils/toCSS/toCSS.spec.ts
+++ b/packages/core/src/utils/toCSS/toCSS.spec.ts
@@ -1,9 +1,5 @@
-import { describe, expect, it, jest } from '@jest/globals';
+import { describe, expect, it } from '@jest/globals';
 import { toCSS } from './toCSS';
-
-jest.mock('../../utils', () => ({
-  kebabCase: (s: string) => s.replace(/([a-z])([A-Z])/g, '$1-$2').toLowerCase(),
-}));
 
 describe('toCSS', () => {
   it('should transform a style object correctly to CSS', () => {
--- a/jest.config.js
+++ b/jest.config.js
@@ -16,6 +16,10 @@ export default {
       statements: 100,
     },
   },
+  moduleNameMapper: {
+    '^@onfido/castor$': '<rootDir>/core/src',
+    '^@onfido/castor-(.*)$': '<rootDir>/$1/src',
+  },
   preset: 'ts-jest',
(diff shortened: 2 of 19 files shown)


packages/core/src/utils/toCSS/toCSS.spec.ts
```diff
@@ -1,10 +1,6 @@
-import { describe, expect, it, jest } from '@jest/globals';
+import { describe, expect, it } from '@jest/globals';
 import { toCSS } from './toCSS';
 
-jest.mock('../../utils', () => ({
-  kebabCase: (s: string) => s.replace(/([a-z])([A-Z])/g, '$1-$2').toLowerCase(),
-}));
-
```
rabelloo, line 7: By removing the mock we're testing `kebabCase` twice, once here and another in its own unit tests.

Which is fine, but this file is no longer strictly "unit" testing.

I'm personally ok with it in this instance but might be trickier in other cases.

rabelloo (approved):
In general I dislike "magic" import resolutions, but I can see how it makes this all easier to parse.

I'll see how I like it going forward.

packages/core/src/utils/toCSS/toCSS.spec.ts
josokinas: I agree that mocking is great in cases when you mock the output directly.

However, here it's mocking the actual utility.

Which IMO does not make sense - so same utility is written twice...

josokinas:
Purpose of this is to align, as we now had both cases used.

Happy to discuss possibility of the other approach, when all paths are relative that point to the same package. As long as it's aligned on every usage.

packages/core/src/utils/toCSS/toCSS.spec.ts
rabelloo: We could rewrite the mock per test, so it doesn't mock the implementation, but the return.

But again, should be fine as is in this one case.

rabelloo:
No worries, the absolute import looks good here, simplifies a lot.

LGTM

Issue #118, opened 2020-12-18
Bundle size tracking


rabelloo:
Include bundle size checks in our CI, like [bundlesize](https://github.com/siddharthkp/bundlesize) or [bundlewatch](https://github.com/bundlewatch/bundlewatch),
so that we keep Castor under a certain size threshold.

josokinas:
Maybe we could also do the same for https://github.com/onfido/castor-icons 🤔

josokinas:
And also for https://github.com/onfido/castor-tokens now 😏

PR #699, opened 2021-05-21
build: check bundle size

kopijunkie:
## Purpose

Add bundle size checking for #118 

## Approach

Add bundle size checking for castor/core and castor/react packages' build files using `bundlewatch`. This check is done as part of build Github Action workflow but can also be done locally by running `yarn check:bundle-size`.

## Testing

- Checked locally by tweaking `maxSize` in the `bundlewatch` config in `package.json` to deliberately trigger a failure when file size is above limit.
- PR build should also fail if this config's `maxSize` is changed to a value below the current sizes.

## Risks

N/A (might be worth defining what maximum file sizes to start with, current limit is set based on rounding up current sizes for each package's files)

### DIFF
--- a/.bundlewatchrc.json
+++ b/.bundlewatchrc.json
@@ -1,0 +1,17 @@
+{
+  "trackBranches": ["master"],
+  "files": [
+    {
+      "path": "./packages/core/dist/index.js",
+      "maxSize": "200B"
+    },
+    {
+      "path": "./packages/core/dist/castor.css",
+      "maxSize": "3kB"
+    },
+    {
+      "path": "./packages/react/dist/index.js",
+      "maxSize": "90B"
+    }
+  ]
+}
--- a/.bundlewatchrc.json
+++ b/.bundlewatchrc.json
@@ -1,5 +1,5 @@
 {
-  "trackBranches": ["master"],
+  "trackBranches": ["main"],
   "files": [
     {
       "path": "./packages/core/dist/index.js",
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,6 +25,11 @@ jobs:
       - name: Build packages
         run: yarn build
 
+      - name: Bundle size check
+        uses: jackyef/bundlewatch-gh-action@0.2.0
+        with:
+          bundlewatch-github-token: ${{ secrets.BUNDLEWATCH_GITHUB_TOKEN }}
+
       - name: Build (bundled) examples
         run: |
           yarn --cwd ./examples/custom-themes build
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,7 +29,7 @@ jobs:
         uses: jackyef/bundlewatch-gh-action@0.2.0
         with:
           bundlewatch-github-token: ${{ secrets.BUNDLEWATCH_GITHUB_TOKEN }}
-          bundlewatch-config: ../../.bundlewatchrc.json
+          bundlewatch-config: .bundlewatchrc.json
 
       - name: Build (bundled) examples
         run: |
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,6 +29,7 @@ jobs:
         uses: jackyef/bundlewatch-gh-action@0.2.0
         with:
           bundlewatch-github-token: ${{ secrets.BUNDLEWATCH_GITHUB_TOKEN }}
+          bundlewatch-config: ../../.bundlewatchrc.json
 
       - name: Build (bundled) examples
         run: |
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -83,5 +85,24 @@
...
+      },
+      {
+        "path": "./packages/core/dist/castor.css",
+        "maxSize": "3kB"
+      },
+      {
+        "path": "./packages/react/dist/index.js",
+        "maxSize": "90B"
+      }
+    ]
```
rabelloo, line 106: 💅🏻 [nitpick]

I'd personally prefer a dedicated `.bundlewatch.json` file for these configs.

@josokinas do you have any preferences?

rabelloo (approved):
I imagine we'll be tweaking the threshold quite frequently as we add new components, so it should be fine to start off with a conservative number.

package.json
josokinas: Yes, I would prefer a separate file, however I'd call it `.bundlewatchrc.json`.

Downside is that seems like you then need to specify that file for CLI: https://bundlewatch.io/#/getting-started/using-a-config-file

package.json
```diff
@@ -37,7 +37,8 @@
     "e2e:run": "wait-on http://localhost:6006 && cypress run -c video=true",
     "e2e:serve": "NODE_ENV=e2e yarn start --ci --quiet",
     "snapshot": "CYPRESS_updateSnapshots=true yarn e2e",
-    "release": "HUSKY=0 standard-version"
+    "release": "HUSKY=0 standard-version",
+    "check:bundle-size": "yarn build && yarn run bundlewatch"
```
josokinas, line 41: I think that because we currently have only one check, it is fine to omit the prefix:

```suggestion
    "check-bundle-size": "yarn build && yarn run bundlewatch"
```

josokinas (approved):

josokinas:
I've renamed this PR to have "build" scope. We cannot have this as "feat" as it would then add to a CHANGELOG, which we would not need to have with this.

PR should be merged with a comment `build: check bundle size` too.

package.json
kopijunkie: Happy to move the config out to a separate file. But yeah, the only disadvantage is it needs to be specified in the CLI command. The Github Action can be easily set up to take the config file path also.

kopijunkie:
> I imagine we'll be tweaking the threshold quite frequently as we add new components, so it should be fine to start off with a conservative number.

Yeah, I've deliberately gone for a more conservative number rather than deciding on a set of arbitrary numbers for now. Definitely can imagine it will need to be bumped up as work on Castor progresses until work to be done stabilises.

.github/workflows/build.yml
```diff
@@ -25,6 +25,12 @@ jobs:
       - name: Build packages
         run: yarn build
 
+      - name: Check bundle size
+        uses: jackyef/bundlewatch-gh-action@0.2.0
+        with:
+          bundlewatch-github-token: ${{ secrets.BUNDLEWATCH_GITHUB_TOKEN }}
+          bundlewatch-config: ../../.bundlewatchrc.json
```
josokinas, line 32: @rabelloo do you prefer this over config in `package.json`?

package.json
```diff
@@ -37,7 +37,8 @@
     "e2e:run": "wait-on http://localhost:6006 && cypress run -c video=true",
     "e2e:serve": "NODE_ENV=e2e yarn start --ci --quiet",
     "snapshot": "CYPRESS_updateSnapshots=true yarn e2e",
-    "release": "HUSKY=0 standard-version"
+    "release": "HUSKY=0 standard-version",
+    "check-bundle-size": "yarn build && yarn run bundlewatch --config ./.bundlewatchrc.json"
```
josokinas, line 41: [Minor] I think can be just:

```suggestion
    "check-bundle-size": "yarn build && yarn run bundlewatch --config .bundlewatchrc.json"
```

.github/workflows/build.yml
josokinas: P.S. This is breaking on CI, maybe just `.bundlewatchrc.json` works?

package.json
kopijunkie: Good point. Meant to double check that this also works then forgot about it.

.github/workflows/build.yml
kopijunkie: Yeap. I forgot that it looks it up from the root of the project, not where the workflow is located. Should be working fine now.

rabelloo: yup this looks great 💪🏻

.bundlewatchrc.json
```diff
@@ -0,0 +1,17 @@
+{
+  "trackBranches": ["master"],
```
rabelloo, line 2: Should be `main`, right?

kopijunkie: Oops. Thought I'd updated that. 🙈

PR #704, opened 2021-05-24
docs: always connect label with input/textarea

josokinas:
## Purpose

Input and Textarea should always be connected to FieldLabel, even when it is nested - due to accessibility (screen readers).

## Approach

In Storybook stories connect Inputs and Textareas to FieldLabels.

## Testing

On Storybook, both Core and React stories of Input/Textarea with FieldLabel (and HelperText).

## Risks

N/A

### DIFF
--- a/packages/core/src/components/input/input.stories.ts
+++ b/packages/core/src/components/input/input.stories.ts
@@ -55,11 +55,13 @@ export const Disabled = htmlMatrix(Input, { disabled });
 Disabled.argTypes = omit<InputProps>('disabled');
 
 interface InputWithLabelAndHelperTextProps extends InputProps {
+  id: string;
   label: string;
   helperText: string;
 }
 
 export const WithLabelAndHelperText = ({
+  id,
   label,
   helperText,
   ...props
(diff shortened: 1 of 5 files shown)


packages/core/src/components/input/input.stories.ts
```diff
@@ -55,23 +55,31 @@ export const Disabled = htmlMatrix(Input, { disabled });
...
   ...props
 }: InputWithLabelAndHelperTextProps) =>
   Field({
     children: [
       FieldLabel({
-        children: [label, HelperText({ children: helperText }), Input(props)],
+        children: [
+          label,
+          HelperText({ children: helperText }),
+          Input({ ...props, id }),
```
rabelloo, line 75: I don't think this line needs to change if we don't destructure the `id` 🤔

rabelloo: Ah nvm you're using it for `for` later on, better than `props.id` I suppose.

packages/react/src/components/form/form.stories.tsx
```diff
@@ -35,35 +35,35 @@ interface Values {
...
     </Field>
 
     <Field>
-      <FieldLabel>
+      <FieldLabel htmlFor="name-name">
         <span>
           Last name <Asterisk aria-label="required" />
         </span>
-        <Input name="lastName" required />
+        <Input id="name-name" name="lastName" required />
```
rabelloo, line 54: Did you mean `last-name`? 😅

rabelloo (approved):

packages/react/src/components/form/form.stories.tsx
josokinas: Yeah, mistake...

packages/core/src/components/input/input.stories.ts
josokinas: Yes, I do not like using `props.<something>` that much...

rabelloo: Me neither

PR #705, opened 2021-05-24
feat(react): auto-generate ids for checkbox/radio + label connections

josokinas:
## Purpose

Due to accessibility each `<label>` should be connected with `<input>` via ID.

## Approach

Provide auto-generated IDs for Checkbox and Radio components (via custom hook).

ID set via prop still takes priority.

## Testing

Unit tested hook; inspected DOM on React stories.

## Risks

N/A


packages/react/src/components/checkbox/checkbox.tsx
```diff
@@ -17,11 +19,14 @@ export const Checkbox = withRef(
     ref: CheckboxProps['ref']
   ): JSX.Element => {
     const { disabled, touched } = useField();
+    const autoId = useAutoId('castor_checkbox');
+    const id = externalId || autoId;
```
rabelloo, line 23: We should probably use a Nullish Coalescing operator instead so that `0` is a valid `id`.

Also, if they do `id=""` I suppose we let them override the `autoId` too?

```suggestion
    const id = externalId ?? autoId;
```

rabelloo: Actually even `id={null}` should be able to override the `autoId`, I think.

Maybe we just default the property value?

```ts
let count = 0;

const Checkbox = ({ id = `castor_checkbox_${++count}` }) => // ...
```

packages/react/src/hooks/useAutoId/useAutoId.spec.ts
```diff
@@ -0,0 +1,25 @@
+import { describe, expect, it } from '@jest/globals';
+import { renderHook } from '@testing-library/react-hooks';
+import { useAutoId } from './useAutoId';
+
+describe('useAutoId', () => {
+  it('should generate with same unique prefix', () => {
+    const { result: result1 } = renderHook(() => useAutoId('castor_test'));
```
rabelloo, line 7: I don't see why this dependency is needed here, `useAutoId` and hooks in general are just pure functions, should be easy to mock and unit test.

```ts
jest.mock('react', () => ({ useMemo: (fn) => fn(), }));

it('should be easy to mock', () => {
  const autoId = useAutoId('castor_test');

  expect(autoId).toBe('castor_test_1');
});
```

packages/react/src/components/checkbox/checkbox.tsx
josokinas: Originally it was [implemented like such](https://github.com/onfido/castor/pull/267/commits/65d002f248780398fad931277d2860c3847f6b76), however was changed to [internal state management](https://github.com/onfido/castor/pull/267/commits/bde7d3361d3306f6ca8dac591bb2abf03a0d2270).

TBH I can't remember why, and this seems to be fine... Unless I'm missing something.

rabelloo: The count is going to increment on every render rather than on every component instance.

There shouldn't be any practical difference, other than the attributes changing a lot more.

rabelloo (approved):

PR #714, opened 2021-05-25
feat: icon requires aria-hidden or aria-label

### LINKED ISSUE #707: Require either aria-hidden or aria-label for icons
When using `Icon` component, add type to require either setting `aria-hidden="true"` or `aria-label`.

rabelloo:
## Purpose

Closes #707

## Approach

Type `IconProps` to require either `aria-hidden` or `aria-label`

## Testing

Icon stories.

## Risks

Breaking change: current integrators will need to retro-apply either attribute to all `<Icon>` instances.

I don't like this risk, I'd introduce the attributes as optional for now.


docs/helpers/aria.ts
```diff
@@ -0,0 +1,22 @@
+import { ArgType } from './story';
+
+const hidden: ArgType<unknown> = {
+  control: 'boolean',
+  description:
+    'Indicates whether the element is exposed to an accessibility API.\n\n' +
+    'Either this or `aria-label` required for accessibility',
```
josokinas, line 7: [Minor]

What if:

```suggestion
    [
      'Indicates whether the element is exposed to an accessibility API.',
      'Either this or `aria-label` required for accessibility',
    ].join('\n\n'),
```

packages/core/src/components/icon/icon.stories.ts
```diff
@@ -15,19 +15,28 @@ export default {
...
+    'aria-label': aria.label,
   },
   args: {
     name: firstIconName,
   },
 } as Meta<IconProps>;
 
 export const Playground: Story<IconProps> = (props) => Icon(props);
+Playground.args = {
+  'aria-label': 'a label for my icon',
```
josokinas, line 29: Just a suggestion:

```suggestion
  'aria-label': 'A label for an icon',
```

packages/core/src/components/icon/icon.stories.ts
```diff
@@ -15,19 +15,28 @@ export default {
         type: { summary: colors.map((value) => `"${value}"`).join('|') },
       },
     },
+    name: {},
```
josokinas, line 18: What does `name` do here?

packages/core/src/components/icon/icon.ts
```diff
@@ -1,7 +1,10 @@
...
 
-export interface IconProps {
+export type IconProps = (Hidden | Labelled) & {
   name: IconName;
   color?: Color;
-}
+};
+
+type Hidden = { 'aria-hidden': true; 'aria-label'?: never };
+type Labelled = { 'aria-hidden'?: never; 'aria-label': string };
```
josokinas, line 10: Because we follow American English in code:

```suggestion
type Labeled = { 'aria-hidden'?: never; 'aria-label': string };
```

Source: https://www.grammarly.com/blog/labeled-labelled/

packages/core/src/components/icon/icon.story.tsx
```diff
@@ -1,14 +1,12 @@
...
-export interface IconProps extends BaseProps {
-  'aria-hidden'?: 'true' | 'false';
-}
-
 /**
  * `.ods-icon` requires SVG sprite to be included in your app.
  *
  * https://github.com/onfido/castor-icons
+ *
+ * Requires either `aria-hidden` or `aria-label` for accessibility.
```
josokinas, line 9: Should it be:

```suggestion
 * Requires either `aria-hidden="true"` or `aria-label` for accessibility.
```

What do you think?

packages/core/src/components/icon/icon.stories.ts
rabelloo: Simply reorders it above `aria-` attributes

packages/core/src/components/icon/icon.ts
rabelloo: TIL they're different

packages/core/src/components/icon/icon.story.tsx
rabelloo: The value is going to be defined by the type, which actually makes me realise it's wrong for non-react, can't be a boolean.

I was just listing the name of the attributes in the comment, it does make sense to show an example, but then should it be `aria-label=""` too?

docs/helpers/aria.ts
```diff
@@ -0,0 +1,23 @@
...
+  ].join('\n\n'),
+  table: { type: { summary: `'true' | 'false'` } },
+};
+
+const label: ArgType<unknown> = {
+  control: 'text',
+  description:
+    'Defines a value that labels the current element.\n\n' +
+    'Either this or `aria-hidden` required for accessibility',
```
josokinas, line 16: Following previous accepted suggestion:

```suggestion
    [
      'Defines a value that labels the current element.',
      'Either this or `aria-hidden` required for accessibility',
    ].join('\n\n'),
```

packages/react/src/components/icon/icon.stories.tsx
```diff
@@ -15,20 +22,29 @@ export default {
...
+    'aria-label': aria.label,
   },
   args: {
     name: firstIconName,
   },
 } as Meta<IconProps>;
 
 export const Playground: Story<IconProps> = (props) => <Icon {...props} />;
+Playground.args = {
+  'aria-label': 'a label for my icon',
```
josokinas, line 36: Following previous accepted suggestion:

```suggestion
  'aria-label': 'A label for an icon',
```

packages/core/src/components/icon/icon.story.tsx
josokinas: Well no, because for `aria-hidden` a requirement is specifically for it to be set as `"true"`, but not `"false"`.

For `aria-label` value does not matter, as long as it is set.

packages/core/src/components/icon/icon.stories.ts
josokinas: I see.

What if it's:

```suggestion
    name: {
      table: { type: { summary: 'string' } },
    },
```

Just less confusing IMO...

rabelloo: What about this? 😅 

```suggestion
    name: {}, // just reorders in the table
```

packages/core/src/components/icon/icon.story.tsx
rabelloo: Fair enough

docs/helpers/aria.ts
rabelloo: 🙈  missed it

rabelloo:
What do you think about the Risk?

docs/helpers/aria.ts
```diff
@@ -0,0 +1,24 @@
+import { ArgType } from './story';
+
+const hidden: ArgType<unknown> = {
+  control: 'boolean',
+  description: [
+    'Indicates whether the element is exposed to an accessibility API.',
+    'Either this or `aria-label` required for accessibility',
```
josokinas, line 7: [Minor] Should this (and the one below) have a full stop?

packages/core/src/components/icon/icon.stories.ts
josokinas: Fine, also TBH can also miss the ordering, I don't mind it.

packages/react/src/components/icon/icon.tsx
```diff
@@ -7,6 +7,8 @@ import React from 'react';
  * `@onfido/castor-icons` instead.
  *
  * https://github.com/onfido/castor-icons
+ *
+ * Requires either `aria-hidden` or `aria-label` for accessibility.
```
josokinas, line 11:
```suggestion
 * Requires either `aria-hidden="true"` or `aria-label` for accessibility.
```

josokinas (approved):
Looks good, build is failing though?

docs/helpers/aria.ts
rabelloo: Yep, nice catch.

packages/core/src/components/icon/icon.stories.ts
rabelloo: No ordering is weird, `color -> aria-hidden -> aria-label -> name`.

Arguably, `name` is the most important.

I'll leave as is with the comment for now, we can review if/when we do MDX.

rabelloo:
If it defaults to something, it's no longer required 🙈 

It won't be any different than having no extra types at all.

josokinas:
Right, it would not.

Then it would be - hiding unless a label is set.

Maybe that's also fine?

rabelloo:
As discussed I'll leave this PR open as a breaking change being introduced, for us to merge later on when we're ready for 2.0

josokinas:
Yes, that's fine 👍

However, if you could still take a look at failing build please 😏

PR #718, opened 2021-05-26
build: lint ts

rabelloo:
## Purpose

Catch TS (type) errors ahead of time, especially in our examples, docs, and test files.

Also make prettier ignore files it doesn't know, like `.hbs`.

## Approach

Run `tsc --noEmit` on lint.

## Testing

Should output some logs in our pipeline.
Run `tsc --noEmit` on `main` and some errors will be found, fixed in this branch.

## Risks

None.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -30,5 +30,8 @@ jobs:
       - name: Run ESLint
         run: yarn lint:es
 
+      - name: Run tsc (type check)
+        run: yarn lint:ts
+
       - name: Run Prettier
         run: yarn lint:prettier
--- a/e2e/tsconfig.json
+++ b/e2e/tsconfig.json
@@ -1,6 +1,7 @@
 {
   "extends": "../tsconfig.json",
   "include": ["**/*"],
+  "exclude": [],
   "compilerOptions": {
     "noEmit": true,
     "sourceMap": false,
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "lint": "concurrently yarn:lint:*",
     "lint:commit": "commitlint --from origin/main --to HEAD --verbose",
     "lint:es": "eslint . --ignore-path .gitignore --max-warnings 0",
+    "lint:ts": "tsc --noEmit",
     "lint:style": "stylelint '**/*.{,s}css' --ignore-path .gitignore --max-warnings 0",
     "lint:prettier": "prettier . --check --ignore-path .gitignore",
     "test": "jest --coverage",
--- a/docs/helpers/html.ts
+++ b/docs/helpers/html.ts
@@ -12,7 +12,7 @@ export const html = (
     `<${name}`,
     Object.entries(props).map(toAttribute).join('\n'),
     '>',
-    children?.join?.('\n') || children,
+    (children as unknown[])?.join?.('\n') || children,
     `</${name}>`,
   ]
     .filter(Boolean)
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,6 +1,6 @@
 {
   "include": ["**/*"],
-  "exclude": ["node_modules/**/*"],
+  "exclude": ["node_modules/**/*", "e2e/**/*"],
   "compilerOptions": {
     "alwaysStrict": true,
     "declaration": false,


package.json
```diff
@@ -28,8 +28,9 @@
     "lint": "concurrently yarn:lint:*",
     "lint:commit": "commitlint --from origin/main --to HEAD --verbose",
     "lint:es": "eslint . --ignore-path .gitignore --max-warnings 0",
+    "lint:ts": "tsc --noEmit",
     "lint:style": "stylelint '**/*.{,s}css' --ignore-path .gitignore --max-warnings 0",
-    "lint:prettier": "prettier . --check --ignore-path .gitignore",
+    "lint:prettier": "prettier . --check --ignore-unknown --ignore-path .gitignore",
```
josokinas, line 33: The thing is, e.g. https://github.com/onfido/castor/pull/683 is failing because Prettier did add support for Handlerars in [v2.3.0](https://prettier.io/blog/2021/05/09/2.3.0.html), so I believe it won't be ignore after that one is merged.

It is not merged in yet, as it is failing, and I'm not looking at that because I'm in a process of removing those files.

rabelloo: Hmm good to know. We should probably run Prettier 2.3.0 on `castor-tokens`' handlebars files then.

I'll revert here.

e2e/tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "../tsconfig.json",
   "include": ["**/*"],
+  "exclude": [],
```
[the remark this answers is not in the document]
rabelloo: It includes all files inside its root, which is now `e2e`.
However the root `tsconfig` excludes `e2e/**/*`.
So we need to reset the excludes array 🤷🏻‍♂️

package.json
josokinas: It is upgraded over there, however I did not find a solution, and instead just [ignored the files](https://github.com/onfido/castor-tokens/blob/cca5fcb04113cc9f1e74dcf12eded6fff59cd5bb/.prettierignore#L2).

Prettier cannot have multiple `--ignore-path`, so if custom is used, the default one (`.prettierignore`) is no longer used, and vice versa)

e2e/tsconfig.json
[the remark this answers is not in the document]
josokinas: Little odd, but yeah fine... 😏

package.json
rabelloo: We probably shouldn't ignore the file then 🤔

e2e/tsconfig.json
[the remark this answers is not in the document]
rabelloo: They behave as two different workspaces though, so if you are in `~/castor` TS will see all files except `e2e`.
If you `cd` into `~/castor/e2e` however, TS will only see files inside that folder, and nothing else.
The config is a bit odd I suppose, because we want to reuse most of the definitions, but the result makes sense I think.

package.json
josokinas: I don't like how it formats it though, which then results in Sass file generated being with even stranger format...

package.json
```diff
@@ -28,6 +28,7 @@
     "lint": "concurrently yarn:lint:*",
     "lint:commit": "commitlint --from origin/main --to HEAD --verbose",
     "lint:es": "eslint . --ignore-path .gitignore --max-warnings 0",
+    "lint:ts": "tsc --noEmit",
```
josokinas, line 31: Should this also [run on CI](https://github.com/onfido/castor/blob/fdc2bcb6920b0fe49656fb40a1af40fc5f787797/.github/workflows/lint.yml#L6)?

josokinas (approved):
Looks good, please if you could adjust PR title and description, as now Prettier unknown files are not excluded 😏

package.json
rabelloo: Ah I see. We could include a comment in the beginning of each `.hbs` file `{{! prettier-ignore }}`

Dunno if it's worth the hassle, `.prettierignore` is fine.
The only disadvantage of using a `.prettierignore` is forgetting to update it when `.gitignore` changes, which is rare.

package.json
rabelloo: Definitely, I forget we don't do `yarn lint` to run all things. Will update.

package.json
josokinas: I tried inline ignore, it did not work... The only thing that worked was `.prettierignore`...

.github/workflows/lint.yml
```diff
@@ -30,5 +30,8 @@ jobs:
       - name: Run ESLint
         run: yarn lint:es
 
+      - name: Run tsc (type check)
```
josokinas, line 33: [Minor] Don't mind either, but maybe:

```suggestion
      - name: Run TypeScript compiler (type check)
```

PR #722, opened 2021-05-27
docs(core): allow children in story to be null or undefined

josokinas:
## Purpose

When building core story for a component and "children" prop is in use, it's type should also accept `null` and `undefined`.

## Approach

Improve "children" prop type to also accept `null` and `undefined`, leaving `string` or `string[]` variation untouched.

## Testing

N/A

## Risks

N/A


rabelloo:
The `html` helper makes all properties effectively optional, is there a reason we'd want `null` over `undefined`?

josokinas:
Yes, it conflicts with story building when you explicitly set certain prop to `null`, or avoid setting at all (`undefined`).

rabelloo (approved):
Cool, `undefined` should be covered by the optional nature of `html` props, but `null` is not.

This is more accurate anyways.

PR #723, opened 2021-05-27
fix(react): hide icon within component using aria

josokinas:
## Purpose

Following https://github.com/onfido/castor/pull/703 icons used in other components should also be hidden from screen readers.

## Approach

Hide icons in Search and Validation (React) components using `aria-hidden="true"`.

## Testing

N/A

## Risks

N/A

### DIFF
--- a/packages/react/src/components/search/search.tsx
+++ b/packages/react/src/components/search/search.tsx
@@ -16,7 +16,7 @@ export const Search = withRef(
   ): JSX.Element => (
     <div className={classy(c('search'), className)} style={style}>
       <Input {...restProps} ref={ref} type="search" />
-      <Icon name="search" />
+      <Icon name="search" aria-hidden="true" />
     </div>
   )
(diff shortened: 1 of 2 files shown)


josokinas:
Such is also fixed in https://github.com/onfido/castor/pull/714 however that one will only be released with a breaking change, as of this is valid fix.

rabelloo (approved):
We should probably cherry-pick the `lint:ts` changes as well 🤔

josokinas:
What changes you're talking about @rabelloo?

rabelloo:
These https://github.com/onfido/castor/pull/718. My bad, I already extracted them 🙃

josokinas:
Right, I'll come back to https://github.com/onfido/castor/pull/718 later, there's some other bits that will be covered in other awaiting PRs too (e.g. no more handlebar files will be in the repo).

PR #724, opened 2021-05-27
build: fix bundlewatch diff report

josokinas:
## Purpose

Following https://github.com/onfido/castor/pull/699 fix bundle size diff.

## Approach

Tracking all `dist` JS and CSS files (of all packages) and prototypes, also setting `main` branch as base and the one to track.

In addition, setting max file sizes (subject to be reviewed), renaming script as "inspect".

## Testing

Report on PR.

## Risks

Not 100% sure diff will work, will need to be merged first in order to take effect.

### DIFF
--- a/.bundlewatchrc.json
+++ b/.bundlewatchrc.json
@@ -1,17 +1,20 @@
 {
-  "trackBranches": ["main"],
+  "ci": {
+    "repoBranchBase": "main",
+    "trackBranches": ["main"]
+  },
   "files": [
     {
-      "path": "./packages/core/dist/index.js",
-      "maxSize": "200B"
+      "path": "./packages/*/dist/**/*.css",
+      "maxSize": "5kB"
     },
     {
-      "path": "./packages/core/dist/castor.css",
-      "maxSize": "3kB"
+      "path": "./packages/*/dist/**/*.js",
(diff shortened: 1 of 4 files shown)


josokinas:
Due to an issue with Bundlewatch [erroring on large file number watch when generating URL](https://github.com/bundlewatch/bundlewatch/issues/34) I'm shortening list of files to only base CSS and prototypes.

.bundlewatchrc.json
```diff
@@ -1,17 +1,16 @@
...
+  },
   "files": [
-    {
-      "path": "./packages/core/dist/index.js",
-      "maxSize": "200B"
-    },
     {
       "path": "./packages/core/dist/castor.css",
-      "maxSize": "3kB"
+      "maxSize": "5KB"
```
rabelloo, line 9: Should we watch for a theme as well?

rabelloo (approved):

.bundlewatchrc.json
josokinas: Theme consists of theme tokens, same as for Castor Tokens package I don't think we have much control over it.

rabelloo: Fair enough.

PR #737, opened 2021-06-03
ci: setup node version based on .nvmrc file

josokinas:
## Purpose

When upgrading Node version, we need to change it in multiple places. However, as we do use `.nvmrc` config, better to use such on CI also.

## Approach

On CI, read Node version required from `.nvmrc` file, then use that when setting up Node.

## Testing

On pipeline:

Screenshot 2021-06-03 at 11 10 53

## Risks

N/A

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,10 +14,14 @@ jobs:
       - name: Checkout Git repository
         uses: actions/checkout@v2.3.4
 
-      - name: Setup Node
+      - name: Read .nvmrc
+        run: echo ::set-output name=NVMRC::$(cat .nvmrc)
+        id: nvm
+
+      - name: Setup Node (uses version from .nvmrc)
         uses: actions/setup-node@v2.1.5
         with:
-          node-version: 14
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -14,10 +14,14 @@ jobs:
...
-      - name: Setup Node
+      - name: Read .nvmrc
+        run: echo ::set-output name=NVMRC::$(cat .nvmrc)
+        id: nvm
+
+      - name: Setup Node (uses version from .nvmrc)
         uses: actions/setup-node@v2.1.5
         with:
-          node-version: 14
+          node-version: ${{ steps.nvm.outputs.NVMRC }}
```
kopijunkie, line 24: Just a minor point, but I feel NVMRC_VERSION would be a better name since the file itself is named `.nvmrc`

kopijunkie (approved):

.github/workflows/build.yml
josokinas: The `.nvmrc` file only holds one value - it's a number version of Node to use... So sort of IMO is clear when defining such as variable also... 😏

PR #747, opened 2021-06-08
docs(react): include select (native) component in form example

josokinas:
## Purpose

Add example with using Select (native) component with a Form.

## Approach

Use Select (native) component to define gender in a form.

## Testing

On Storybook, React story of Form component.

## Risks

N/A

### DIFF
--- a/packages/react/src/components/form/form.stories.tsx
+++ b/packages/react/src/components/form/form.stories.tsx
@@ -10,7 +10,9 @@ import {
   FormProps,
   HelperText,
   Input,
+  Option,
   Radio,
+  Select,
   Validation,
 } from '@onfido/castor-react';
 import React from 'react';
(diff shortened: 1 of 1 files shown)


packages/react/src/components/form/form.stories.tsx
```diff
@@ -89,6 +92,23 @@ export const Playground: Story<FormProps<Values>> = (props) => (
...
+    <Field>
+      <FieldLabel htmlFor="gender">
+        <span>
+          Gender <Asterisk />
+        </span>
+        <Select id="gender" name="gender" native required>
+          <Option value="">Select an option...</Option>
+          <Option value="m">Male</Option>
+          <Option value="f">Female</Option>
+          <Option value="n/a">Prefer not to answer</Option>
```
rabelloo, line 104: Let's add a third option to be a bit more inclusive

```suggestion
          <Option value="n-b">Non-binary</Option>
          <Option value="n/a">Prefer not to answer</Option>
```

rabelloo (approved):

PR #752, opened 2021-06-09
feat: add select (custom) component

josokinas:
## Purpose

Add Select component.

## Approach

Implements Select  (custom) component.

This closes issue #3.

## Testing

On Storybook preview.

## Risks

N/A


packages/core/src/components/select/select.scss
```diff
@@ -53,26 +75,60 @@
         color: helpers.color('content-disabled');
       }
     }
+  }
 
-    &.-borderless {
-      @include mixins.with-inner-focus('action');
-      @include with-padding(); // no border width because border will be unset
+  .ods-select-custom.-open + .ods-icon {
+    transform: rotate(180deg);
```
rabelloo, line 81: 💅🏻  [nitpick]

We should rotate a single axis, small perf boost, either Y or Z I believe.

packages/core/src/components/select/select.scss
```diff
@@ -53,26 +75,60 @@
...
+  }
 
-      background-color: transparent;
-      border: 0;
+  .ods-select-custom-dropdown {
+    @include mixins.with-inner-focus('action');
```
rabelloo, line 85: The "border" being our general "on focus inset box-shadow" can be a problem:

image

Granted, the dropdown should probably not be visible in the case when it's not focused.

So although it makes it DRY to reuse our mixin, I believe it works better as an actual border here.

Also, I wonder if a tiny "box-shadow" for a sense of elevation would be good.

packages/core/src/components/select/select.scss
```diff
@@ -53,26 +75,60 @@
...
 
-      &:hover {
-        background-color: helpers.color('background-action-subtle-hover');
-      }
+    background-color: helpers.color('background-input');
+    border-radius: helpers.border-radius('small');
+    box-sizing: border-box;
+    display: none;
+    margin-top: helpers.space(0.5);
+    padding: 4px; // spread set on `with-inner-focus` mixin
```
rabelloo, line 92:
```suggestion
    padding: helpers.space(0.5); // spread set on `with-inner-focus` mixin
```

packages/core/src/components/select/select.scss
```diff
@@ -53,26 +75,60 @@
...
+    background-color: transparent;
+    border: 0;
+    color: helpers.color('content-main');
+    display: flex;
+    outline: none;
+    padding: helpers.space(1.5) helpers.space(2);
+    text-align: left;
+    width: 100%;
+
+    &.-focused {
```
rabelloo, line 120: Can this not be `:focus`?

packages/react/src/components/select/custom/custom-option.tsx
```diff
@@ -0,0 +1,75 @@
...
+    changeOption,
+    removeOption,
+  } = useCustomSelect();
+
+  const value = externalValue.toString();
+  const focused = currentFocusOption?.id === internalId;
+  const selected = currentValue === value;
+
+  useEffect(() => {
+    addOption(internalId, { value, title });
```
rabelloo, line 32: Since the `ref` is already kept, the key could be that object's hash instead of a custom id we keep track of.

packages/react/src/components/select/custom/custom-select-portal.tsx
```diff
@@ -0,0 +1,116 @@
...
+  }, [id]);
+
+  // evaluate lazily, so that this runs only once and ref points to same element
+  // https://reactjs.org/docs/hooks-faq.html#how-to-create-expensive-objects-lazily
+  const getRootElem = () => {
+    if (!ref.current) ref.current = document.createElement('div');
+    return ref.current;
+  };
+
+  return getRootElem();
```
rabelloo, line 65: Am I missing something here or is the wrapping function not needed at all?

```suggestion
  return ref.current ??= document.createElement('div');
```

packages/react/src/components/select/custom/custom-select-portal.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+function createElement(id: string): HTMLDivElement {
+  const element = document.createElement('div');
+  element.setAttribute('id', id);
+  return element;
+}
+
+function addElement(el: Element): void {
+  document.body.appendChild(el);
+}
```
rabelloo, line 76: I am not sure we should be creating the container outside of React's control, perhaps asking the user to "save a slot" for the Select dropdown container would be better?

I mean, the current way makes it much more convenient, I'm just worried we might run into other sorts of bugs, especially in MFE scenarios.

rabelloo:
You weren't kidding when you said there was a lot of JS 😅 

I still wonder if there's a way to get most of what we need from vanilla elements so we don't need so much custom code, but that is probably a better concern for a second iteration of the component.

packages/core/src/components/select/select.scss
josokinas: It cannot, because during keyboard navigation a class is moved from item to item.

That is, a state keeps which option should be focused on, then option itself adds `.-focused` if it's the one that should be focused on.

packages/react/src/components/select/custom/custom-select-portal.tsx
josokinas: This now adds a container for portal if not found, created with static ID "castor_custom_select_dropdown".

So, to avoid such, you'd only need to make sure a container with this ID is in the document.

We could treat this as an edge case, which would involve that extra step. Or even document as a solution.

Otherwise we can omit creating a container component, but still require such to be added manually when using a component.

We could also provide e.g. a CastorPortal component that is responsible to create all required containers, for Select and other components (like Tooltip if needed).

packages/core/src/components/select/select.scss
josokinas: The spread is a hardcoded value, based on requirements and calculations within the mixin(s).

I cannot use a space here, because changing a base would mess up the component.

I will however add an extra note to better identify that this must be a hardcoded value.

packages/core/src/components/select/select.scss
josokinas: Probably that's the case - I will move from using a mixin to just adding to a dropdown element itself.

I do not want to add elevation here, only because elevation tokens are to be still added by @Subcide, so we can add at a later stage as component improvement.

packages/react/src/components/select/select.stories.tsx
```diff
@@ -13,6 +13,8 @@ import { Meta, omit, reactMatrix, Story } from '../../../../../docs';
 const borderless = [true, false] as const;
 const disabled = [true, false] as const;
 const invalid = [true, false] as const;
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const native = [true, false] as any;
```
josokinas, line 17: @rabelloo feel free to suggest an alternative here, I could not find a solution to this type issue.

packages/react/src/components/select/custom/custom-option.tsx
josokinas: I tried this, and it would work.

However, because I also use that ID as `key` when iterating over options and building `<option>`, I cannot use `ref` for that...

packages/core/src/components/select/select.scss
rabelloo: Interesting. I don't think I get the full picture but I'll continue to explore the component later.

packages/react/src/components/select/select.stories.tsx
rabelloo: `as const` doesn't work?

josokinas: No, it does not.

That one is a little different, it has `true` described on native component props, and `false` on custom one.

packages/react/src/components/select/custom/useDropdownKeyboardKey.spec.ts
```diff
@@ -0,0 +1,539 @@
...
+  haltEvent: () => void
+) => void;
+
+jest.mock('../../../hooks', () => ({
+  useKeyboardKey: jest
+    .fn()
+    .mockImplementation((_, callback: any) => (callbackFunc = callback)),
+}));
+
+describe('useDropdownKeyboardKey', () => {
```
josokinas, line 17: Don't know even what to think about this test, over 500 lines... 😟

packages/react/src/components/select/custom/custom-select-portal.tsx
rabelloo: Yea if that's the way we choose to go a component like `CastorPortal` would be very nice.

That's a breaking change, so might be good to choose now.

What do you think, should we require a Portal target like we do `Icons` for the sheet?

packages/react/src/components/select/custom/custom-option.tsx
rabelloo: I see. That's fine then, not a big deal.

packages/react/src/components/select/select.stories.tsx
rabelloo: `as boolean[]` doesn't work either?
What's the expected type?

packages/react/src/components/select/custom/useDropdownKeyboardKey.spec.ts
rabelloo: I know right, which is why I'm always wary of introducing JS to do something the DOM can natively.

As I said in the other comment, we can iterate and refactor this component without API changes in the near future, hopefully cutting a lot of the custom code.

packages/react/src/components/select/custom/custom-select-portal.tsx
josokinas: Probably that would be best. Same pattern.

However, note that next release will already most likely be a 2.0 so we can do a breaking change.

packages/react/src/components/select/select.stories.tsx
josokinas: Tried that, tried different variations - nothing worked.

It says type is `readonly true[] | readonly false[]` (or something like this), however if such set then there is another error...

packages/react/src/components/select/custom/custom-select-portal.tsx
rabelloo: Right. Since we're not releasing "CustomSelect" until next release, we can work on that Portal component then.

packages/react/src/components/select/select.stories.tsx
rabelloo: It's probably an issue in `reactMatrix`'s type, I can take a look later, `as any` is fine for now.
`as any[]` should probably work too.

josokinas:
Closing in favour of https://github.com/onfido/castor/pull/965.

PR #766, opened 2021-06-16
feat: add progress component

euljr:
## Purpose

Implement the `Progress` component

## Approach

Create the new component using `div` with `role="progressbar"` instead of the HTML5 `progress` element, the reason is because the `progress` element requires several workarounds for the CSS to be consistent between browsers.

It also sets the recommended `aria` attributes, more info:

- https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_progressbar_role
- https://www.w3.org/TR/wai-aria-1.1/#progressbar

Other Progress Bars also use this approach, for example: 

- https://baseweb.design/components/progress-bar/
- https://luna.sainsburys.co.uk/components/progress-indicator

## Testing

Storybook `progress.story.tsx`

## Risks

N/A


packages/react/src/components/progress/progress.tsx
```diff
@@ -0,0 +1,38 @@
...
+      {!hideLabel && (
+        <div className={classy(c('progress-label'), m(size))}>
+          {children || `${normalizedValue}%`}
+        </div>
+      )}
+    </div>
+  );
+};
+
+export type ProgressProps = BaseProps & JSX.IntrinsicElements['progress'];
```
josokinas, line 38: The element used is actually not a `<progress>` but a `<div>`. That's why props of a JSX should also be of that component:

```suggestion
export type ProgressProps = BaseProps & JSX.IntrinsicElements['div'];
```

Then, I would exclude all the types that are overridden in the component. When setting such on TypeScript you'd get an error, which is better then putting something in but getting it rewritten.

```suggestion
export type ProgressProps = BaseProps & Omit<JSX.IntrinsicElements['div'], 'aria-valuemax' | 'aria-valuemin' | 'aria-valuenow' | 'aria-valuetext' | 'role'>;
```

P.S. The last one will fail, as it will need to be run against Prettier if you accept the suggestion.

packages/react/src/components/progress/progress.tsx
```diff
@@ -0,0 +1,38 @@
+import { c, classy, m, ProgressProps as BaseProps } from '@onfido/castor';
+import React from 'react';
+
+export const Progress = ({
+  value = 0,
+  max = 100,
+  size = 'regular',
+  hideLabel,
```
josokinas, line 8: I wonder if we should follow `hide` instead of `show`? That is, should this rather be `showLabel` that defaults to `true`?

This would also make JSX (IMO) nicer, where you do `showLabel && ...` instead of `!hideLabel && ...`.

@rabelloo what do you think?

packages/react/src/components/progress/progress.tsx
```diff
@@ -0,0 +1,38 @@
...
+  children,
+  className,
```
josokinas, line 10: You also need to pass the rest of props to the parent.

So:

```suggestion
  className,
  ...restProps
```

And then follow the next comment.

packages/react/src/components/progress/progress.tsx
```diff
@@ -0,0 +1,38 @@
...
+}: ProgressProps): JSX.Element => {
+  const normalizedValue = Math.round((value / max) * 100);
+
+  return (
+    <div
+      className={classy(c('progress'), m(size), className)}
```
josokinas, line 16: We usually spread the props first, then override what's needed.

```suggestion
    <div
      {...restProps}
      className={classy(c('progress'), m(size), className)}
```

packages/react/src/components/progress/progress.tsx
```diff
@@ -0,0 +1,38 @@
...
+      role="progressbar"
+      aria-valuenow={value}
+      aria-valuemin={0}
```
josokinas, line 19: I wonder, do we have a usecase to also set a `min`?

For example, what if your progress is a part of 50 to a 100, so 50 would be 0%?

packages/core/src/components/progress/progress.ts
```diff
@@ -0,0 +1,7 @@
+export interface ProgressProps {
+  children?: string;
```
josokinas, line 2: No need to set `children`, it's a part of element props.

```suggestion
```

josokinas (commented):
I understand this is not finalised yet, I did only a quick component review for now.

packages/react/src/components/progress/progress.tsx
euljr: I'm fine also with `showLabel` instead of `hide`, I agree that `show` wold make the JSX nicer, but I think `hide` has a better syntax for the user:

`<Progress value={50} hideLabel />`

instead of

`<Progress value={50} showLabel={false} />`

I'm fine with either, so let me know what you think and I change it.

josokinas: That's also a valid point.

Probably let's leave this for now as it is.

packages/react/src/components/progress/progress.tsx
euljr: Yes, I think that's exactly what happens, some implementations like [this](https://github.com/uber/baseweb/blob/master/src/progress-bar/progressbar.js#L99) sets it default to 0, but I don't see a problem in having the option.

I can default it to 0, what do you think?

josokinas: I think yes to be more future-proof, better to have this as optional prop defaulting to a `0`.

packages/core/src/components/progress/progress.ts
euljr: I have to set it as string though, if not, typescript complains for the `aria-valuetext` prop.

[About the `aria-valuetext`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_progressbar_role#example_2_using_aria-valuetext)

There are other options I can think of:
 - Remove the `aria-valuetext` and let this up to the end user to add
 - Use a string prop `label` instead of `children`

What do you think is best?

packages/core/src/components/progress/progress.scss
```diff
@@ -0,0 +1,57 @@
...
+    width: 100%;
+
+    &.-regular {
+      height: helpers.space(0.5);
+    }
+
+    &.-large {
+      height: helpers.space(1);
+    }
+  }
```
rabelloo, line 38: We can simplify the interface to a single element by using pseudo-elements and CSS vars.

image

`:before` is the background, `:after` is the progressed amount, and by applying text styles directly to `.ods-progress` we don't need the inner `label`.

Makes the vanilla usage much simpler

```html
<div
  class="ods-progress -regular"
  role="progressbar"
  aria-valuenow="25"
  aria-valuemin="0"
  aria-valuemax="100"
  style="--progress: 25%"
>
  25%
</div>
```

packages/core/src/components/progress/progress.story.tsx
```diff
@@ -0,0 +1,37 @@
...
+  ...props
+}: ProgressProps) =>
+  html('div', {
+    ...props,
+    class: classy(c('progress'), m(size)),
+    role: 'progressbar',
+    'aria-valuenow': String(value),
+    'aria-valuemin': String(0),
+    'aria-valuemax': String(max),
+    'aria-valuetext': !hideLabel && children ? children : undefined,
```
rabelloo, line 19: 💅🏻  [nitpick]

```suggestion
    'aria-valuetext': (!hideLabel && children) || undefined,
```

packages/react/src/components/progress/progress.stories.tsx
```diff
@@ -0,0 +1,83 @@
...
+};
+
+export const WithoutLabel = (props: ProgressProps) => <Progress {...props} />;
+WithoutLabel.argTypes = omit<ProgressProps>('children');
+WithoutLabel.args = {
+  children: null,
+  hideLabel: true,
+};
+
+export const Example = (props: ProgressProps) => {
```
rabelloo, line 70: Can we figure out a better name? Arguably every story is an `Example` 🙈 
Perhaps `Progression` or `Value` or `ValueLoop`?

I also noticed the code shown does not update, because it's internal state.
I wonder if there's a way to update the story's props and in turn the shown code.
Perhaps there's a `Storybook controls` hook?

packages/react/src/components/progress/progress.stories.tsx
```diff
@@ -0,0 +1,83 @@
...
+  const [value, setValue] = useState(0);
+
+  useEffect(() => {
+    setTimeout(() => {
+      setValue(value === 100 ? 0 : value + 10);
+    }, 1000);
```
rabelloo, line 76: 💅🏻  [nitpick]

```suggestion
    setTimeout(setValue, 1000, value === 100 ? 0 : value + 10);
```

packages/react/src/components/progress/progress.tsx
rabelloo: I would prefer if no label was the default and `<Progress autoLabel />` would generate an automatic label out of the value.

However the requirement is that the auto-label is "on" by default, so...

I prefer `<Progress value={50} hideLabel />` too

packages/react/src/components/progress/progress.tsx
rabelloo: That's so other props apply, like `data-` and `id` and so on.

packages/react/src/components/progress/progress.tsx
rabelloo: I'd say implement it when we need it as it's not a breaking change, but I don't see why not do it now as it seems to be simple.

packages/react/src/components/progress/progress.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+export const Progress = ({
+  value = 0,
+  max = 100,
+  size = 'regular',
+  hideLabel,
+  children,
+  className,
+}: ProgressProps): JSX.Element => {
+  const normalizedValue = Math.round((value / max) * 100);
```
rabelloo, line 12: 💅🏻  [nitpick]

What do y'all think of `percentValue` instead?

packages/core/src/components/progress/progress.ts
rabelloo: I like `children`, although it's a React convention I don't mind other (future) rendering engines calling it children either.
We can always change the interface specific to a single engine if needed.

I'm fine with setting it to `string` here.

rabelloo: Actually, I take that back, that's no good, we lose the ability to compose, say:

```tsx
<Progress value={22} value={40}>
  Step <MyCode>22</MyCode> / <MyCode>40</MyCode>
</Progress>
```

Let's not define it as `string`, rather we will need more runtime code in React's component, e.g.

```tsx
  aria-valuetext={typeof children === 'string' ? children : undefined}
```

It'd be ideal if we could read the `textContent` of `children` that are `JSX.Element`s but that might require too much code, with no real benefit.

packages/react/src/components/progress/progress.tsx
euljr: Done!

packages/react/src/components/progress/progress.tsx
euljr: Done!

packages/react/src/components/progress/progress.tsx
euljr: Done!

packages/react/src/components/progress/progress.tsx
euljr: I have added the min prop.

packages/core/src/components/progress/progress.ts
euljr: I have changed it to:

```
aria-valuetext={
        ariaValuetext || (typeof children === 'string' ? children : undefined)
}
```
So the user can also provide the `aria-valuetext` himself, what do you think?

packages/react/src/components/progress/progress.stories.tsx
euljr: I was thinking if I shouldn't remove this altogether, we don't have any "running" example for other components.

It's nice to see it working, but the "show code" only make sense if we can see the context around (like the useEffect hook, etc).

Trying to find a way to do that, but maybe I will remove it, what do you think?

packages/core/src/components/progress/progress.ts
rabelloo: Yeap that looks good.

Lazy question because I don't feel like reading the entire w3 document again, can't it infer the `valuetext` from the `textContent` of the element?

packages/react/src/components/progress/progress.stories.tsx
rabelloo: We mostly do static examples because of core/vanilla.

I like the idea of a dynamic example but would like the code to update alongside state.
For which we need to use props, I think.
Decorators might do it too.

I don't know, maybe it's not worth the effort right now, but it's definitely nice to have.

packages/core/src/components/progress/progress.scss
euljr: Way better! Can you and also @josokinas take a look in the new implementation?

packages/react/src/components/progress/progress.stories.tsx
euljr: I have added the example manually. Can you take a look? It isn't perfect as we have to change both when updating the test, but at least, when clicking on `copy` the user will get a "working" example, missing the imports only.

Let me know if this is OK, if isn't, I think I will remove it for now.
If you think is best to remove, I
In case you think this is a problem, I think

packages/core/src/components/progress/progress.scss
```diff
@@ -0,0 +1,54 @@
...
+      background-color: helpers.color('background-progress');
+      border-radius: helpers.border-radius('full');
+      content: '';
+      width: 100%;
+    }
+
+    &::after {
+      background-color: helpers.color('content-progress');
+      border-radius: helpers.border-radius('full');
+      content: '';
```
rabelloo, line 28: 💅🏻  [nitpick]

```suggestion
    &::before,
    &::after {
      border-radius: helpers.border-radius('full');
      content: '';
    }
    
    &::before {
      background-color: helpers.color('background-progress');
      width: 100%;
    }

    &::after {
      background-color: helpers.color('content-progress');
```

rabelloo (approved):

packages/core/src/components/progress/progress.ts
euljr: I think we have to fill it when we change the default (which is a percentage number).

> Authors SHOULD only set the `aria-valuetext` attribute when the rendered value cannot be meaningfully represented as a number.

https://www.w3.org/TR/wai-aria-1.1/#aria-valuetext

packages/core/src/components/progress/progress.scss
rabelloo: Looks neat!

rabelloo:
Pre approving as all other things are minor to me, did you mean to mark it as "non-draft"?

packages/core/src/components/progress/progress.ts
rabelloo: Cool so it'd read `valuenow` incorrectly if the `valuemax` is different. Let's leave it then.

packages/core/src/components/progress/progress.story.tsx
euljr: This is changed now to match the new implementation so it's not exactly as you have suggested.

packages/react/src/components/progress/progress.tsx
```diff
@@ -0,0 +1,46 @@
...
+      {...restProps}
+      className={classy(c('progress'), m(size), className)}
+      role="progressbar"
+      aria-valuenow={value}
+      aria-valuemin={min}
+      aria-valuemax={max}
+      aria-valuetext={
+        ariaValuetext || (typeof children === 'string' ? children : undefined)
+      }
+      style={{ ...style, ...cssVars({ percentValue }) }}
```
josokinas, line 35: This will break if `style = null`.

I would suggest to not default to `{}` upon prop destruction, but instead:

```suggestion
      style={{ ...(style ?? {}), ...cssVars({ percentValue }) }}
```

packages/react/src/components/progress/progress.tsx
```diff
@@ -0,0 +1,46 @@
...
+  max = 100,
+  size = 'regular',
+  style = {},
+  hideLabel,
+  children,
+  className,
+  'aria-valuetext': ariaValuetext,
+  ...restProps
+}: ProgressProps): JSX.Element => {
+  const percentValue = `${Math.round(((value - min) * 100) / (max - min))}%`;
```
josokinas, line 22: What about [memoizing value](https://reactjs.org/docs/hooks-reference.html#usememo) here:

```suggestion
  const percentValue = useMemo(() => `${Math.round(((value - min) * 100) / (max - min))}%`, [value, min, max]);
```

packages/core/src/components/progress/progress.scss
```diff
@@ -0,0 +1,54 @@
...
+@mixin Progress() {
+  @if not mixins.includes('Progress') {
+    @include _Progress();
+  }
+}
+
+@mixin _Progress() {
+  .ods-progress {
+    display: inline-grid;
+    grid: auto-flow/auto;
```
josokinas, line 13: [Minor]

```suggestion
    grid: auto-flow / auto;
```

packages/core/src/components/progress/progress.stories.ts
```diff
@@ -0,0 +1,71 @@
+import { htmlMatrix, Meta, omit, Story } from '../../../../../docs';
+import { Progress, ProgressProps } from './progress.story';
+
+const size = ['regular', 'large'] as const;
+
+export default {
+  title: 'Core/Progress',
+  component: Progress,
+  argTypes: {
```
josokinas, line 9: This now has unneeded `aria-valuetext` in table, better to:

```suggestion
  argTypes: {
    ...omit<ProgressProps>('aria-valuetext'),
```

packages/core/src/components/progress/progress.stories.ts
```diff
@@ -0,0 +1,71 @@
...
+        type: { summary: size.toString() },
+        defaultValue: { summary: 'text' },
+      },
+    },
+    hideLabel: {
+      description: 'Hide label',
+      table: {
+        defaultValue: { summary: 'false' },
+      },
+    },
```
josokinas, line 43: No need to describe some obvious props:

```suggestion
    value: {
      table: {
        defaultValue: { summary: '0' },
      },
    },
    min: {
      table: {
        defaultValue: { summary: '0' },
      },
    },
    max: {
      table: {
        defaultValue: { summary: '100' },
      },
    },
    size: {
      control: { type: 'radio', options: size },
      table: {
        type: { summary: size.toString() },
        defaultValue: { summary: 'text' },
      },
    },
    hideLabel: {
      table: {
        defaultValue: { summary: 'false' },
      },
    },
```

packages/react/src/components/progress/progress.stories.tsx
```diff
@@ -0,0 +1,119 @@
...
+        type: { summary: size.toString() },
+        defaultValue: { summary: 'text' },
+      },
+    },
+    hideLabel: {
+      description: 'Hide label',
+      table: {
+        defaultValue: { summary: 'false' },
+      },
+    },
```
josokinas, line 46: Same here:

```suggestion
    value: {
      table: {
        defaultValue: { summary: '0' },
      },
    },
    min: {
      table: {
        defaultValue: { summary: '0' },
      },
    },
    max: {
      table: {
        defaultValue: { summary: '100' },
      },
    },
    size: {
      control: { type: 'radio', options: size },
      table: {
        type: { summary: size.toString() },
        defaultValue: { summary: 'text' },
      },
    },
    hideLabel: {
      table: {
        defaultValue: { summary: 'false' },
      },
    },
```

packages/core/src/components/progress/progress.story.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+export interface ProgressProps extends BaseProps {
+  children?: string | null;
+  'aria-valuetext'?: string;
+}
+
+export const Progress = ({
+  size,
+  value,
+  max = 100,
```
josokinas, line 12: `min` missing:

```suggestion
  min = 0,
  max = 100,
```

packages/core/src/components/progress/progress.stories.ts
```diff
@@ -0,0 +1,71 @@
...
+  },
+  args: {
+    children: '',
+    hideLabel: false,
+    max: 100,
+    size: 'regular',
```
josokinas, line 49: Should these default?

```suggestion
```

packages/react/src/components/progress/progress.stories.tsx
```diff
@@ -0,0 +1,119 @@
...
+  },
+  args: {
+    children: '',
+    hideLabel: false,
+    max: 100,
+    size: 'regular',
```
josokinas, line 52: So also these:

```suggestion
```

packages/core/src/components/progress/progress.story.tsx
```diff
@@ -0,0 +1,33 @@
...
+  html('div', {
+    ...props,
+    class: classy(c('progress'), m(size)),
+    role: 'progressbar',
+    'aria-valuenow': String(value),
+    'aria-valuemin': String(0),
+    'aria-valuemax': String(max),
+    'aria-valuetext':
+      ariaValuetext || (typeof children === 'string' ? children : undefined),
+    style: `--percent-value: ${Math.round((value / max) * 100)}%`,
```
josokinas, line 27: Maybe new utility could be used?

Also, I believe calculation is not up to date with the React version here.

packages/core/src/utils/cssVars/cssVars.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * Generates a plain object of CSS variable key/value pairs based on the specified `vars`.
+ * @note Filters out falsy values and transforms names to kebab-case.
+ * @example
```
josokinas, line 4: Minor style changes:

```suggestion
 * Generates a plain object of CSS variable key/value pairs based on the
 * specified `vars`.
 *
 * @note Filters out falsy values and transforms names to kebab-case.
 *
 * @example
```

packages/core/src/components/progress/progress.stories.ts
```diff
@@ -0,0 +1,71 @@
...
+WithoutLabel.args = {
+  children: undefined,
+  hideLabel: true,
+};
+
+export const CustomLabel: Story<ProgressProps> = (props) => Progress(props);
+CustomLabel.argTypes = omit<ProgressProps>('children');
+CustomLabel.args = {
+  children: 'Progress: 25%',
+};
```
josokinas, line 71: [Minor] To match React stories, flipping places:

```suggestion
export const CustomLabel: Story<ProgressProps> = (props) => Progress(props);
CustomLabel.argTypes = omit<ProgressProps>('children');
CustomLabel.args = {
  children: 'Progress: 25%',
};

export const WithoutLabel: Story<ProgressProps> = (props) => Progress(props);
WithoutLabel.argTypes = omit<ProgressProps>('children');
WithoutLabel.args = {
  children: undefined,
  hideLabel: true,
};
```

packages/react/src/components/progress/progress.stories.tsx
```diff
@@ -0,0 +1,119 @@
...
+};
+
+export const WithoutLabel = (props: ProgressProps) => <Progress {...props} />;
+WithoutLabel.argTypes = omit<ProgressProps>('children');
+WithoutLabel.args = {
+  children: null,
+  hideLabel: true,
+};
+
+export const AllCombinations = reactMatrix(
```
josokinas, line 78: I would suggest to make this a last story, because it is not used for humans but for automated testing.

Also, I believe CSS version is missing "All Combinations"

josokinas (approved):

packages/react/src/components/progress/progress.stories.tsx
rabelloo: On Storybook it looks good, before AllCombinations is better though, that should always be the last story.

On code, it looks... well, repetitive. Does converting the function to a string work?

```ts
docs: {
  code: ValueLoop.toString()
}
```

packages/core/src/components/progress/progress.story.tsx
euljr: Nice catch! Yes! An utility would be nice in both `core` where we have to set the values manually, and `react` when we want to set a custom label but also show the percentage.

I'm thinking in something like `toProgressPercentValue(min, max, value): string`, what do you think?

Also, not sure where to put it 🤔

josokinas: Sorry, I just meant the `cssVars` utility...

euljr: Ah sorry! Lost the context a bit in the morning 🙈 

About using the `cssVars` here, different from the `react` code, the `html` helper is only to generate a helper right?

If this is the case, I don't see an advantage of using here, for two reason:
 
 - The cssVars generate react objects, like `{ percentageValue: '40%' }`, we would have to map it to strings
 - Simply setting the string may be more readable than going through function descriptions in order to understand what's happening
 
 So I think it's doable, but it would require more code here or another utility. What do you think?

packages/react/src/components/progress/progress.stories.tsx
euljr: I agree, converting the function to string gets the JS instead of the JSX code:

```
function ValueLoop(props) {
  var _useState = Object(react__WEBPACK_IMPORTED_MODULE_16__["useState"])(0),
      _useState2 = _slicedToArray(_useState, 2),
      value = _useState2[0],
      setValue = _useState2[1];

  Object(react__WEBPACK_IMPORTED_MODULE_16__["useEffect"])(function () {
    setTimeout(setValue, 1000, value === 100 ? 0 : value + 10);
  }, [value]);
  return /*#__PURE__*/Object(react_jsx_runtime__WEBPACK_IMPORTED_MODULE_18__["jsx"])("div", {
    "aria-busy": value !== 100,
    style: {
      width: '100%'
    },
    children: /*#__PURE__*/Object(react_jsx_runtime__WEBPACK_IMPORTED_MODULE_18__["jsx"])(_onfido_castor_react__WEBPACK_IMPORTED_MODULE_15__[/* Progress */ "a"], Object.assign({}, props, {
      value: value
    }))
  });
}
```

Maybe remove the example for now? I can try to find better solutions some other time.

packages/core/src/components/progress/progress.story.tsx
josokinas: That's fine, don't worry.

I guess it's only a matter of aligning with React so it's using both `min` and `max`.

packages/react/src/components/progress/progress.tsx
```diff
@@ -0,0 +1,38 @@
...
+  const normalizedValue = Math.round((value / max) * 100);
+
+  return (
+    <div
+      className={classy(c('progress'), m(size), className)}
+      role="progressbar"
+      aria-valuenow={value}
+      aria-valuemin={0}
+      aria-valuemax={max}
+      aria-valuetext={!hideLabel && children ? children : undefined}
```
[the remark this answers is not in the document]
euljr: This has also changed, can you check if it's ok?

packages/core/src/components/progress/progress.scss
euljr: Done!

packages/react/src/components/progress/progress.tsx
euljr: Sure, done!

packages/core/src/components/progress/progress.scss
euljr: Done!

packages/core/src/components/progress/progress.stories.ts
euljr: Nice catch, didn't notice that, changed, thanks!

packages/core/src/components/progress/progress.stories.ts
euljr: Done!

packages/react/src/components/progress/progress.stories.tsx
euljr: Done!

packages/core/src/components/progress/progress.story.tsx
euljr: Added, thanks!

packages/core/src/components/progress/progress.stories.ts
euljr: Removed them, thanks!

packages/react/src/components/progress/progress.stories.tsx
euljr: Removed them, thanks!

packages/core/src/components/progress/progress.story.tsx
euljr: The calculation is fixed here and in another line after that, thanks!

packages/core/src/utils/cssVars/cssVars.ts
euljr: Changed, thanks!

packages/core/src/components/progress/progress.stories.ts
euljr: Done!

packages/react/src/components/progress/progress.stories.tsx
euljr: Done, thanks!

packages/react/src/components/progress/progress.stories.tsx
```diff
@@ -0,0 +1,112 @@
...
+    max: {
+      table: {
+        defaultValue: { summary: '100' },
+      },
+    },
+    size: {
+      control: { type: 'radio', options: size },
+      table: {
+        type: { summary: size.toString() },
+        defaultValue: { summary: 'text' },
```
josokinas, line 35: This displays as default value "text", I believe you want to set here "regular".

packages/core/src/components/progress/progress.stories.ts
```diff
@@ -0,0 +1,77 @@
...
+    max: {
+      table: {
+        defaultValue: { summary: '100' },
+      },
+    },
+    size: {
+      control: { type: 'radio', options: size },
+      table: {
+        type: { summary: size.toString() },
+        defaultValue: { summary: 'text' },
```
josokinas, line 35: ...and same here, it should be "regular" a default.

packages/core/src/components/progress/progress.story.tsx
```diff
@@ -0,0 +1,36 @@
...
+    'aria-valuemin': String(min),
+    'aria-valuemax': String(max),
+    'aria-valuetext':
+      ariaValuetext || (typeof children === 'string' ? children : undefined),
+    style: `--percent-value: ${Math.round(
+      ((value - min) * 100) / (max - min)
+    )}%`,
+    children: [
+      !hideLabel
+        ? children || `${Math.round(((value - min) * 100) / (max - min))}%`
```
josokinas, line 33: [Minor] Up to you, as this is only for Storybook, however because you do calculation twice maybe you could extract this to a `percentValue` constant and share in these two places.

packages/react/src/components/progress/progress.stories.tsx
```diff
@@ -0,0 +1,112 @@
...
+WithoutLabel.args = {
+  children: null,
+  hideLabel: true,
+};
+
+export const CustomLabel = (props: ProgressProps) => <Progress {...props} />;
+CustomLabel.argTypes = omit<ProgressProps>('children');
+CustomLabel.args = {
+  children: 'Progress: 25%',
+};
```
josokinas, line 69: [Minor] To match both stories:

```suggestion
export const CustomLabel = (props: ProgressProps) => <Progress {...props} />;
CustomLabel.argTypes = omit<ProgressProps>('children');
CustomLabel.args = {
  children: 'Progress: 25%',
};

export const WithoutLabel = (props: ProgressProps) => <Progress {...props} />;
WithoutLabel.argTypes = omit<ProgressProps>('children');
WithoutLabel.args = {
  children: null,
  hideLabel: true,
};
```

josokinas:
Final few very minor comments added, otherwise LGTM. Nice job 👍

josokinas:
Actually, last thing - when you finish addressing all the comments, please don't forget to [add new E2E screenshots](https://github.com/onfido/castor/blob/main/CONTRIBUTING.md#testing-locally):

> If you introduced visual changes intentionally and are sure it's how it should look, update the screenshot baselines by downloading `screenshots` artifacts in your PR's checks and copying the images into `./e2e/.snapshots`.

Let me know if you have questions regarding this.

packages/react/src/components/progress/progress.stories.tsx
rabelloo: Ah that's too bad. I don't know, I would personally keep it I think.

@josokinas ?

packages/react/src/components/progress/progress.tsx
rabelloo: That is not true, `null` can be spread just fine, as can any other type.

image

We don't need either.

packages/react/src/components/progress/progress.stories.tsx
josokinas: I like that this way a JS is shown.

However, for such example I don't believe that JS code is needed - it actually just a mockup to make a movement.

I like thought that `aria-busy` is used, so if the example by default renders a JSX and shows both aria and value changed, I would be happy with just it. Otherwise we can live with code duplication.

rabelloo: Not sure I got it. So keep it?

josokinas: Keep it.

rabelloo:
LGTM too, great work on this @euljr!

packages/react/src/components/progress/progress.stories.tsx
euljr: Done, thanks!

packages/core/src/components/progress/progress.stories.ts
euljr: Done, thanks!

packages/core/src/components/progress/progress.story.tsx
euljr: Absolutely! Done.

packages/react/src/components/progress/progress.stories.tsx
euljr: Sorry, should have fixed this before when you commented for the other file, changed, thanks!

PR #775, opened 2021-06-22
docs: remove webpack constant build issue "fix" on storybook

josokinas:
## Purpose

There was [an issue with webpack and Storybook](https://github.com/storybookjs/storybook/issues/14342), that was causing a constant build when running in development mode.

It was "fixed" in https://github.com/onfido/castor/pull/554.

However, now with upcoming Storybook v6.3 [it will be fixed](https://github.com/storybookjs/storybook/issues/14342#issuecomment-852015506), and such is no longer needed.

## Approach

Remove the "fix".

## Testing

Only after Storybook v6.3 upgrade.

## Risks

N/A

### DIFF
--- a/.storybook/main.ts
+++ b/.storybook/main.ts
@@ -13,6 +13,7 @@ module.exports = {
     '@storybook/addon-controls',
     '@storybook/addon-a11y',
   ],
+  stories: ['../{docs,packages}/**/*.stories.ts{,x}'],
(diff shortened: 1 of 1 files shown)


rabelloo (approved):

rabelloo:
We could later experiment with the Storybook Vite builder, should make it much faster
https://storybook.js.org/blog/storybook-for-vite

PR #840, opened 2021-07-13
fix(core): use css properties from csstype, do not rely on react

josokinas:
## Purpose

Because a Font type within a helper uses CSSProperties that come from React dependency, a core package also requires React even for those who do not use it (but use TS).

## Approach

Instead of using CSSProperties that require to have React as a dependency even for a core package, depend directly on `csstype`.

Downside - introducing a dependency.

## Testing

Everything should still work as previously.

## Risks

N/A

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -25,7 +25,8 @@
     "!src/**/*.d.ts"
   ],
   "dependencies": {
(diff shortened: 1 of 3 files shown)


rabelloo (approved):
Can't say I'm thrilled to introduce the dependency but since it's type only it should be alright.
I'll keep an eye.

PR #841, opened 2021-07-13
fix(core): font helper CSS-in-JS compatibility

rabelloo:
## Purpose

Makes `toString` in the object that `font` helper returns non-enumerable so that CSS-in-JS tools won't try to apply it when using object notation, e.g.

```tsx
styled.div(font('300-regular'));
```

## Approach

Define the property in an non-enumerable way with `Object.defineProperty`.

## Testing

Unit tests' objects must match, and therefore no longer have `toString`.

### DIFF
--- a/packages/core/src/helpers/font/font.spec.ts
+++ b/packages/core/src/helpers/font/font.spec.ts
@@ -8,7 +8,6 @@ describe('font', () => {
     fontWeight: 'initial',
     lineHeight: 'initial',
     textTransform: 'none',
-    toString: expect.any(Function),
   };
 
   it('should allow selection by size only', () => {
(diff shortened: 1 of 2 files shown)


rabelloo:
Just a note, `toString() { return toCSS(this); }` also works, CSS-in-JS tools will ignore it, but by making it non-enumerable it's both easier to unit test and it avoids other edge cases we might not be aware of.

posthardcode (approved):

josokinas (approved):
I still cannot replicate this issue on our Emotion example, though...

But happy if that works for an app that is having an issue.

rabelloo:
Here's one

image

image

I've noticed our examples are a bit funky, you need to `npm run build` on the root for them to update `@onfido/castor` code.
Probably something to do with yarn workspaces and symlinks.

But the unit test should make it clear, no `toString` property is present in the object, therefore CSS-in-JS can't mistakenly apply it.

PR #854, opened 2021-07-21
test: run cypress via docker

josokinas:
## Purpose

In order to have alignment between CI and local (Mac) development, we decided to run Cypress via Docker.

## Approach

Run E2E Cypress tests in a Docker container with Compose.

Also use XQuartz to communicate between Cypress in Docker container, and Cypress GUI locally.

On CI, caching is in place for Docker images.

Sources used:
- https://www.cypress.io/blog/2019/05/02/run-cypress-with-a-single-docker-command/
- https://sourabhbajaj.com/blog/2017/02/07/gui-applications-docker-mac/

## Testing

Locally, CI should also pass with newly generated snapshots.

## Risks

This heavily now relies on Mac being a development machine.

However, leaving docs on how to open Cypress GUI locally.


package.json
```diff
@@ -10,7 +10,7 @@
     "node": ">= 14"
   },
   "scripts": {
-    "postinstall": "husky install && yarn integrate",
+    "postinstall": "yarn integrate && if test -d .git; then husky install; fi",
```
josokinas, line 13: This was needed because I do not copy _everything_ to Docker container, and otherwise it would complain "it's not a Git project".

CONTRIBUTING.md
```diff
@@ -86,37 +86,52 @@ They are written in [Cypress](https://www.cypress.io/).
...
 
-Alternatively you could use a single terminal with:
+    IP=$(ipconfig getifaddr en0)
+    /usr/X11/bin/xhost + $IP
 
-    yarn concurrently -k -n ,cypress:open yarn:e2e:serve "yarn cypress open --env failOnSnapshotDiff=false"
+Then open Cypress GUI (using same terminal window that had the IP address set):
 
-For a nicer experience, [you can report snapshot diffs in your test results to console](https://github.com/jaredpalmer/cypress-image-snapshot#reporter):
+    DISPLAY=$IP:0 CYPRESS_baseUrl=http://host.docker.internal:6006 docker-compose -f docker-compose.yml -f docker-compose.open.yml up --abort-on-container-exit --exit-code-from cypress
```
josokinas, line 122: Currently I placed this in docs (full command).

Happy to place within `package.json` scripts, if you think it's better place for that?

rabelloo:
Few questions

> This heavily now relies on Mac being a development machine.

1. From code it seems that only a Bash terminal is needed, so Linux and Windows should work as long as that condition is met, no?
Or is it because XQuartz only works properly on Mac?

2. It seems that XQuartz is only needed for `cypress open`, and `cypress run` works fine inside Docker without external communication, is that right?
If that's the case, do you think it'd be better to avoid the extra setup complexity and just run `cypress open` like before, outside Docker, and either just disable screenshot taking or disregard them, requiring `snapshot` to be run to update baseline images?

3. Do you think it's still valuable to keep the `baseline screenshots` job in CI just in case someone has issues with Docker or for any other reason can't update the images locally?
Might be useful when we're not able to use our work machines.

josokinas:
> 1. From code it seems that only a Bash terminal is needed, so Linux and Windows should work as long as that condition is met, no?
>    Or is it because XQuartz only works properly on Mac?

XQuartz is a Mac software, yes.

This should also work on another system, adding missing pieces. However, docs inform about Mac usage. On the other hand, we already do use Homebrew on docs, so assuming that macOS is used...

> 2. It seems that XQuartz is only needed for `cypress open`, and `cypress run` works fine inside Docker without external communication, is that right?
>    If that's the case, do you think it'd be better to avoid the extra setup complexity and just run `cypress open` like before, outside Docker, and either just disable screenshot taking or disregard them, requiring `snapshot` to be run to update baseline images?

I'm not sure.

I think better to use same machine as used for screenshot taking. That is, Cypress would always run in a Docker container.

What you suggest seems even more confusing - taking screenshots (that is, overriding them), then running locally. But later running via Docker to re-generate proper screenshot.

> 3. Do you think it's still valuable to keep the `baseline screenshots` job in CI just in case someone has issues with Docker or for any other reason can't update the images locally?
>    Might be useful when we're not able to use our work machines.

We could keep it, but only as manual job - otherwise no CI would be longer.

CONTRIBUTING.md
```diff
@@ -86,37 +86,62 @@ They are written in [Cypress](https://www.cypress.io/).
...
+
+For writing tests locally, you would first need to generate snapshots so they are match your environment:
+
+    CYPRESS_updateSnapshots=true yarn concurrently -k --success first -n ,cypress:run yarn:e2e:serve "yarn wait-on http://localhost:6006 && yarn cypress run -c video=true"
 
-Beware that environment differences between CI and your local machine might be enough for a diff to be detected.
+Then open Cypress with:
 
-For writing tests locally:
+    yarn concurrently -k --success first -n ,cypress:open yarn:e2e:serve "yarn cypress open --env failOnSnapshotDiff=false"
```
rabelloo, line 109: Let's make this a script like you suggested before, it's a bit verbose otherwise.

rabelloo (approved):

PR #869, opened 2021-07-26
feat(react): forwardRef all components

rabelloo:
## Purpose

Upcoming component Tooltip will be able to target any element, so we now have an use case for all components to expose their refs.

## Approach

Add `withRef` to all components.

## Testing

- Rendering each component with `ref={ref}` should work with no console logs/errors.
- Each component story's "view code" should still show the component with the correct name.

## Risks

None.

### DIFF
--- a/packages/react/src/components/button/button.tsx
+++ b/packages/react/src/components/button/button.tsx
@@ -1,13 +1,17 @@
 import { ButtonProps as BaseProps, c, classy, m } from '@onfido/castor';
 import { useField } from '@onfido/castor-react';
 import React, { HTMLAttributes } from 'react';
+import { withRef } from '../../utils';
 
-export const Button: ButtonComponent = ({
-  kind = 'action',
-  variant = 'primary',
-  className,
-  ...restProps
-}: ButtonProps<'a'> | ButtonProps<'button'>) => {
+export const Button: ButtonComponent = withRef(function Button(
+  {
+    kind = 'action',
+    variant = 'primary',
+    className,
+    ...restProps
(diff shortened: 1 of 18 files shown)


packages/react/src/components/button/button.tsx
```diff
@@ -1,35 +1,41 @@
...
   const { disabled } = useField();
 
   const Element = (restProps as ButtonProps<'a'>).href ? 'a' : 'button';
 
   return (
     <Element
-      {...(Element === 'button' && {
-        disabled, // will be overriden by props if set
-      })}
+      // will be overriden by props if set
```
josokinas, line 21: [Minor] I prefer the old style of comment, as to me it is more clear what it targets.

josokinas (approved):

josokinas:
One last thing, I believe it's a `feat(react)`.

packages/react/src/components/validation/validation.tsx
```diff
@@ -10,14 +11,17 @@ import React from 'react';
...
+export const Validation = withRef(function Validation(
+  {
+    if: key,
+    state,
+    withIcon,
+    children,
+    className,
+    ...restProps
+  }: ValidationProps,
+  ref?: ValidationProps['ref']
```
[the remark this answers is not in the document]
rabelloo: I don't think it is, e.g. `<Validation />` => `ref === null`, vs `<Validation ref={foo} />` => `ref === foo`

packages/react/src/components/button/button.tsx
rabelloo: Sure, reverted.

PR #878, opened 2021-07-27
feat: implement popover

rabelloo:
## Purpose

Implement Popover, which will be used as a base component in both Select and Tooltip.

## Approach

Core Popover is just a placement container for arbitrary content.
It relies on a `position: relative` parent which contains its target and the Popover itself.

React Popover has the option to use Portals to avoid being hidden by parents' overflow.
If a Portal is used (by setting a `target` ref), it does not require the aforementioned parent anymore.

## Testing

Storybook.

## Risks

Screen aware repositioning coming in a later PR.


rabelloo:
Depends on #869, will wait for it to get merged to `main` then change target.

josokinas:
> Depends on #869, will wait for it to get merged to `main` then change target.

Feel free to merge that one in, already approved.

packages/core/src/components/popover/popover.scss
```diff
@@ -0,0 +1,94 @@
...
+@mixin _positionFrom($side) {
+  #{$side}: 100%;
+  margin-#{$side}: helpers.space(1);
+}
+
+@mixin _center($axis) {
+  $side: if($axis == X, left, top);
+
+  #{$side}: 50%;
+  transform: translate#{$axis }(-50%);
```
josokinas, line 93:
```suggestion
  transform: translate#{$axis}(-50%);
```

josokinas (approved):

packages/core/src/components/popover/popover.scss
rabelloo: So, that is Prettier doing its thing. Try applying that change locally then saving, you'll see 😅 

I suppose we could disable prettier in this line if you really hate how it looks.

josokinas: LOL, never mind... Maybe Prettier will sort itself at some point with some upgrade.

rabelloo:
One other thing I forgot to mention, the `placement` is handled by a single class, e.g. `-top-center`.
It's theoretically two modifiers, one for `position` and another for `alignment`.
But they're treated as a single thing.

For `Button`, we do `.-action--primary` to combine two modifiers.

We could potentially split them, which would make the CSS a bit better and terser, e.g. `-top -center`.
It wouldn't be possible if we did `-top-left` as recommended in design, but it is with `start | end | center`.

Thoughts?

josokinas:
> Thoughts?

I prefer as it is now.

Less freedom of combination - less chance of some edge case.

rabelloo:
Cool. Should we do `.-top--center` as `Button` does?
I can't remember why we chose `--` there.

josokinas:
> Cool. Should we do `.-top--center` as `Button` does?

OK, I see what you mean. Yes, I think better to follow the `Button` component.

rabelloo:
If that's the case, should we:
1. Split them in the component API as well: `<Popover position="top" alignment="center" />`
2. Keep the API but replace the hyphen: `'top-center'.replace('-', '--')`
3. Make the component exactly the same as the class: `<Popover placement="top--center" />`

packages/react/src/components/popover/useOnClickOutside.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+    cleanup();
+
+    expect(removeEventListener).toBeCalledTimes(1);
+  });
+
+  it('should call onClickOutside on outside click', () => {
+    const onClickOutside = jest.fn();
+    const target = { current: { contains: () => false } as any };
+    const event = { target };
+    addEventListener.mockImplementationOnce((_, fn: (_: unknown) => void) =>
```
markorodic, line 55: The underscore convention denotes unused args?

packages/react/src/components/popover/useObserver.ts
```diff
@@ -0,0 +1,31 @@
...
+  const timeout = useRef<number>();
+
+  useEffect(() => {
+    const elements = targets.map((t) => t.current).filter(Boolean) as T[];
+    if (!elements.length) return;
+
+    const observer = new IntersectionObserver(
+      ([entry]) => {
+        clearTimeout(timeout.current);
+        timeout.current = setTimeout(onChange, debounceTime, entry) as never;
```
markorodic, line 23: Not clear why we need a timeout here?

packages/react/src/components/popover/useObserver.ts
```diff
@@ -0,0 +1,31 @@
...
+ * @param targets `RefObject` of element to observe.
+ * @param onChange Called when the threshold changes. Either the element just became visible or the opposite.
+ * @param debounceTime Time in milliseconds to debounce events.
+ */
+export function useObserver<T extends Element>(
+  onChange: (entry: IntersectionObserverEntry) => void,
+  targets: RefObject<T>[],
+  debounceTime?: number
+) {
+  const timeout = useRef<number>();
```
markorodic, line 14: Why are we using a ref here for the timeout rather than some state?

packages/react/src/components/popover/useObserver.ts
```diff
@@ -0,0 +1,31 @@
+import { RefObject, useEffect, useRef } from 'react';
+
+/**
+ * Observes `targets` for intersection changes.
+ * @param targets `RefObject` of element to observe.
+ * @param onChange Called when the threshold changes. Either the element just became visible or the opposite.
+ * @param debounceTime Time in milliseconds to debounce events.
+ */
+export function useObserver<T extends Element>(
```
markorodic, line 9: Is this generic providing flexibility on the type element because all `HTMLElements` inherit from Element? This is a nice use of generics :)

packages/core/src/components/popover/popover.scss
```diff
@@ -0,0 +1,94 @@
...
+
+    &.-left-end,
+    &.-right-end {
+      bottom: 0;
+    }
+    // <========== alignment
+  }
+}
+
+@mixin _positionFrom($side) {
```
markorodic, line 84: I like this solution in the end.

packages/core/src/components/popover/popover.scss
```diff
@@ -0,0 +1,94 @@
...
+    z-index: 9999999;
+  }
+
+  .ods-popover {
+    height: fit-content;
+    max-width: helpers.space(50);
+    position: absolute;
+    width: max-content;
+
+    // position ==========>
```
markorodic, line 24: Is this a common convention for CSS comments?

e2e/visual-regression/all.e2e.ts
```diff
@@ -17,7 +18,14 @@ describe('components', () => {
 
       cy.visit(`/iframe.html?id=${id}`);
 
-      cy.get('body').should('be.visible').matchImageSnapshot(name);
+      // use slightly higher timeout because sometimes the iframe is slow
+      cy.get('body', { timeout: 10000 }).should('have.class', theme);
+
+      // give React one cycle to run effects and timeouts and stuff,
+      // and give the browser one requestFrame
+      cy.wait(20);
```
markorodic, line 26: Wondering whether this test could be more deterministic? Perhaps display a success state when the effects have finished running or the iFrame has loaded.

markorodic (approved):

packages/react/src/components/popover/useOnClickOutside.spec.ts
rabelloo: Yup, avoids `eslint` complaining too. Any name starting with `_` it will ignore.

packages/react/src/components/popover/useObserver.ts
rabelloo: To debounce if set.
For example if `debounceTime === 300` and multiple events trigger in under 300ms, `onChange` will only be called once.
We ended up not using `debounceTime` yet but I suspect I'll need to, in order to fix a bug with the "auto-optimal-positioning" feature coming next.

packages/react/src/components/popover/useObserver.ts
rabelloo: If we used state, every time the `effect` was called we'd update it, triggering a re-run of the entire hook.
Furthermore, that state would need to be part of the `effect`'s dependencies, so we'd have to destroy the previous Observer and create a new one.

Using `ref` makes it so we prevent a lot of render cycles, while still having access to the latest timeout "id" which is all we want.

packages/react/src/components/popover/useObserver.ts
rabelloo: Correct. Observers require an `Element` only so we need not be more specific than that.

In this case the generic doesn't matter because we eat `T` but it would if the return type was T, for instance.
If we did not use the generic then, it'd squash the definition from the outside.

For example:
```ts
const buttonRef = useRef<HTMLButtonElement>(null);

const getCurrentWithGenerics = <T extends Element>(ref: RefObject<T>) =>
  ref.current;
const getCurrentWithPlainType = (ref: RefObject<Element>) => ref.current;

const noiceType = getCurrentWithGenerics(buttonRef);
// HTMLButtonElement | null

const sadType = getCurrentWithPlainType(buttonRef);
// Element | null
```

packages/core/src/components/popover/popover.scss
rabelloo: Thanks for helping me iterate on it, the Sass loop looked horrendous, as did the verbose one-by-one definitions 😂

packages/core/src/components/popover/popover.scss
rabelloo: Nope, that's me. We could wrap it in a mixin too I suppose, or use a different comment style.
Happy to hear suggestions if you have them.
I did not spend a lot of time coming up with this, I just wanted to convey where it begins and where it ends.

e2e/visual-regression/all.e2e.ts
rabelloo: The reason we need this is for Popover's Observer to fire and reposition the anchor element.
I tried thinking of a deterministic way, but it'd be specific to the Popover story, and this spec goes over several other components too.
I couldn't query for an `.ods-popover-anchor` because is not every story that would have one.

Any ideas?

rabelloo: Did some more testing, turns out the theme class is only applied to the iframe's body after one React cycle so I could remove the `cy.wait`.
If it becomes flaky later, we can think of an additional check to perform.

josokinas:
> If that's the case, should we:
> 
> 1. Split them in the component API as well: `<Popover position="top" alignment="center" />`
> 2. Keep the API but replace the hyphen: `'top-center'.replace('-', '--')`
> 3. Make the component API exactly the same as the class: `<Popover placement="top--center" />`
> 4. Use another separation token in the component API:
>    `<Popover placement="top center" />` and `'top center'.replace(' ', '--')`

To best align with a Button component, I would prefer 1st approach.

Others, especially the 4th one, I do not like - e.g. would you expect to also be able to set "center top"? Which in that case would not work.

rabelloo:
> would you expect to also be able to set "center top"?

No, they're typed with the format `${position}-${alignment}`.

The only disadvantage of option 1 is verbosity, but I suppose it makes it more explicit and clearer too, as well as easier to use half of the default, e.g.
`<Popover position="bottom" /> // alignment is default 'center' `

I'm not entirely convinced which option I like best, but option 1 aligns better with Button as you said, so I'm happy to give that a try for version 1.
Will refactor asap.

rabelloo:
[Refactor done](https://github.com/onfido/castor/pull/878/commits/035a4a00f0e9017fa11aa0985882f45eb879e458)

API is now split `<Popover place="bottom" align="start" />`

Let me know if you like it, if so we can merge.

packages/react/src/components/popover/popover.tsx
```diff
@@ -0,0 +1,121 @@
+import {
+  Alignment,
+  c,
+  classy,
+  m,
+  PopoverProps as BaseProps,
+  Position,
+} from '@onfido/castor';
+import React, { ReactNode, RefObject, useRef, useState } from 'react';
+import { createPortal } from 'react-dom';
```
josokinas, line 10: Probably this PR should now include [`react-dom` as dependency](https://github.com/onfido/castor/pull/752/files#diff-1f344ac391eeecc21ec0f01fb07430a47f4b80d20485c125447d54c33c4bbfc4R33)?

packages/react/src/components/popover/popover.tsx
```diff
@@ -0,0 +1,121 @@
...
+import { useObserver } from './useObserver';
+import { useOnClickOutside } from './useOnClickOutside';
+
+export interface PopoverProps extends BaseProps, Omit<Div, 'ref'> {
+  children?: ReactNode;
+  onClose?: () => void;
+  target?: RefObject<Element>;
+}
+
+export const Popover = ({
```
josokinas, line 21: Curious, why did you choose verb-y `align` and `place` rather than noun-y `alignment` and `placement`?

rabelloo: Mostly because it's terser, but also because they exist in CSS like `align-items` and `place-items`.
One could argue that `place` represents `justify + align` but we don't `justify` the popover.

Perhaps `position & align` works better, which is pretty much how Popover relates to CSS.
i.e. `position: top; align-items: start` (even though the implementation doesn't use Flex or Grid).

Happy to go with nouns too.
Naming is hard.

josokinas: Your call on this 😏

packages/react/src/components/popover/popover.tsx
rabelloo: Ah yes, thanks for the reminder.

packages/react/src/components/popover/popover.tsx
rabelloo: Went with `position & align`.
There's still time to change before releasing.

packages/core/src/components/popover/popover.stories.ts
```diff
@@ -0,0 +1,69 @@
...
+export default {
+  title: 'Core/Popover',
+  component: Popover,
+  argTypes: {
+    align: {
+      control: { type: 'inline-radio', options: align },
+      defaultValue: 'center',
+      table: {
+        defaultValue: { summary: 'center' },
+        type: { summary: align.join('|') },
```
josokinas, line 17: Because https://github.com/onfido/castor/pull/882 was merged, please could you switch here (and in other places) to [use `optionsToSummary`](https://github.com/onfido/castor/pull/882/files#diff-28c6eb9281509786c4eec4fd1c77134e899e182aa1abfcd90e8b4168d3fea16dR10).

rabelloo: Done.

josokinas:
Screenshot sizes are different, did you generate them locally via Docker?

I think try to re-generate, as something does not add up.

rabelloo:
No, I downloaded from GitHub Actions, it's weird.

PR #880, opened 2021-07-27
build: use vite for storybook

rabelloo:
## Purpose

Speed up development, `webpack` was taking ~60s to build, `Vite` takes ~1s.

## Approach

Use `storybook-builder-vite` and change some configs.

## Testing

- CI should pass
- `yarn start` should work very much the same (but way faster)

## Risks

Not beta, but still early in Storybook.


packages/core/README.md
```diff
@@ -43,7 +43,7 @@ Choose _one_ theme from the following options:
 `@use` Castor within your main Sass file:
 
 ```scss
-@use '~@onfido/castor';
+@use '@onfido/castor';
```
josokinas, line 46: Do you think such change on docs is needed?

So most likely users will have a setup that is not requiring a `~`?

josokinas (approved):

packages/core/README.md
rabelloo: I was thinking we should remove `~` from everywhere except our examples where actually need them.

It's a requirement of `webpack` and `Parcel 1` but all other bundlers, including `Parcel 2`, can't actually parse the tilde without an alias.

rabelloo: Nevermind, it seems it's still needed, so let's keep it for now.

.storybook/styles.scss
```diff
@@ -1,4 +1,4 @@
-@use '~@onfido/castor';
+@use '@onfido/castor';
```
josokinas, line 1: Did you want to still leave this?

rabelloo: Nope, my mistake, thanks.

rabelloo: Actually this is the only one Vite builds, I think. It breaks without `~`.
https://vercel.com/onfido-oss/castor/76qkGNNpPMGSP4gsaz5rXUSwc8ft

Reverting.

rabelloo:
It breaks all Core stories and the code snippets shown, I'm not sure why.

image

This is a major blocker. I'll try to solve it later but I'm not sure what's changed.

rabelloo:
@josokinas seems that not much has happened on `storybook-builder-vite` so far.

I decided to change the config to use Vite in development (local) for a faster and better DX, but continue to use webpack in production because I couldn't find a fix for the issues described before.

Do you think that's good for now?

josokinas:
Yep, fine, good to go 👍

rabelloo:
Actually it still borks some CSS, story decorators and our custom HTML to React `prepareForInline` function, even locally.

Applied `HOLD` label and will give it more time to stabilise :/

PR #881, opened 2021-07-28
test: mount additional volumes during e2e testing

josokinas:
## Purpose

Docker image needs to be rebuilt each time a code (or config) is changed.

This is because files are copied over when building the image.

## Approach

Instead of copying everything during build, mount additional volumes that are then being synced.

`packages` still need to be copied over during build, because `yarn install` needs to do some actions within these directories.

## Testing

Tested locally, when modifying a file and running snapshot generation.

## Risks

N/A

### DIFF
--- a/Dockerfile.cypress
+++ b/Dockerfile.cypress
@@ -2,12 +2,8 @@ FROM cypress/base:14
 
 WORKDIR /cypress
 
-COPY ./e2e ./e2e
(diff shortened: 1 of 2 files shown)


rabelloo (approved):

rabelloo:
Hope this makes it faster, Docker is by far the slowest bit in the pipeline 😢

PR #882, opened 2021-07-29
docs: improve stories

josokinas:
## Purpose

Improve some Storybook stories.

## Approach

Mainly to make sure defaults are selected, also set on a table (excluding "core" stories, as they should not have defaults).

Also making sure available values to be styled as in other stories - `"<someting>"`.

## Testing

On Storybook.

## Risks

N/A

### DIFF
--- a/docs/helpers/docs.ts
+++ b/docs/helpers/docs.ts
@@ -1,4 +1,16 @@
 /**
+ * Converts options array to a summary string.
+ *
+ * @param options Options array.
+ *
+ * @example
+ * optionsToSummary(['one', 'two'])
+ * // '"one" | "two"'
+ */
+export const optionsToSummary = (options: unknown) =>
+  (options as string[]).map((option) => `"${option}"`).join('|');
+
+/**
  * Creates an object that disables the specified `argTypes`'s controls on the
  * @storybook/docs.
  *
--- a/docs/helpers/docs.ts
+++ b/docs/helpers/docs.ts
@@ -7,8 +7,8 @@
  * optionsToSummary(['one', 'two'])
  * // '"one" | "two"'
  */
-export const optionsToSummary = (options: unknown) =>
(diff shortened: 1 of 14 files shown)


packages/react/src/components/progress/progress.stories.tsx
```diff
@@ -3,6 +3,8 @@ import React, { useEffect, useState } from 'react';
 import { Meta, omit, reactMatrix, Story } from '../../../../../docs';
 
 const size = ['regular', 'large'] as const;
+size.toString = () => size.map((value) => `"${value}"`).join('|');
```
rabelloo, line 6: This stringifier function could be a helper, since we need it in several stories.

rabelloo (approved):

packages/react/src/components/progress/progress.stories.tsx
josokinas: Did create a helper for this, please confirm (last commit) if it's OK, or you'd suggest something different.

docs/helpers/docs.ts
```diff
@@ -1,3 +1,15 @@
...
+ * Converts options array to a summary string.
+ *
+ * @param options Options array.
+ *
+ * @example
+ * optionsToSummary(['one', 'two'])
+ * // '"one" | "two"'
+ */
+export const optionsToSummary = (options: unknown) =>
+  (options as string[]).map((option) => `"${option}"`).join('|');
```
rabelloo, line 11: Can't we type the input?

```suggestion
export const optionsToSummary = (options: readonly string[]) =>
  options.map((option) => `"${option}"`).join('|');
```

packages/react/src/components/progress/progress.stories.tsx
rabelloo: Looks good, left a comment there.

PR #883, opened 2021-07-29
feat: implement tooltip

### LINKED ISSUE #465: Tooltip
Implement the Tooltip component.

Guidelines: https://www.uiguideline.com/components/tooltip

rabelloo:
## Purpose

Closes #465

## Approach

Reuse Popover for placement, apply styles on top on the same element.

## Testing

New stories added for both Core and React.

## Risks

None.


packages/core/src/components/popover/popover.scss
```diff
@@ -1,6 +1,5 @@
 @use '../../helpers';
 @use '../../mixins';
-@use 'sass:map';
```
josokinas, line 3: Would be good for linting to catch this, but I know earlier I was looking into this and did not find a out-of-the-box solution.

josokinas (approved):

packages/core/src/components/popover/popover.scss
rabelloo: Yea automate all the things!
I missed this in the other PR, as well as several changes to Popover's stories, could transfer the changes over if you think that's necessary.

rabelloo: Moved some Popover changes over, should make this PR more concise and targeted.

rabelloo:
Merging. If there are any notes afterwards I can open a new PR to address them.

PR #901, opened 2021-08-06
docs: omit popover class

rabelloo:
## Purpose

Omit Popover's class in its core story.

The second commit is just a refactor that simplifies types for Stories and the `omit` helper used in them.

## Approach

Add `omit('class')`.
Look for `Story` and `omit` instances to simplify their types.

## Testing

`class` should no longer appear in Core/Popover story table.
Everything else should be the same.

## Risks

None.


josokinas (approved):
Still not a fan of PR having other stuff, would like always prefer it to be another PR 😅

rabelloo:
Want me to split it into another PR?
Otherwise I can commit without squashing.

josokinas:
Your call.

It's approved 👍

PR #925, opened 2021-08-17
fix(react): popover with portal reposition on document resize

rabelloo:
## Purpose

Reposition portal'ed Popover on `document.body` resize.

Before
before

After
after

## Approach

Use `ResizeObserver` on `document.body` and reposition Popover on trigger.
Refactored `useObserver` for reusability, and moved hooks to `./utils`.

## Testing

Popover or Tooltip stories, in React, with Portal.

## Risks

None.


packages/react/src/components/popover/popover.tsx
```diff
@@ -68,13 +71,16 @@ function PopoverWithPortal({
...
   Required<Pick<PopoverProps, 'align' | 'position' | 'target'>>) {
+  const domBody = useRef(document.body);
   const popover = useRef<HTMLDivElement>(null);
   const [placement, setPlacement] = useState([position, align] as const);
   const [anchor, setAnchor] = useState(at(target));
 
   useOnClickOutside(onClose, [target, popover]);
 
-  useObserver(
+  useResizeObserver(() => setAnchor(at(target)), [domBody]);
```
rabelloo, line 81: This is the main fix line, the rest is implementing `useResizeObserver` and refactor/cleanup.

packages/react/src/utils/useDebounce/useDebounce.ts
```diff
@@ -0,0 +1,19 @@
+import { useRef } from 'react';
+
+/**
+ * Delays `callback` execution until there are no calls in `debounceTime`.
+ * @param callback Callback to execute.
+ * @param debounceTime Time in milliseconds to debounce events.
+ * @returns
```
josokinas, line 7: Why this `@returns` here?

josokinas (approved):

packages/react/src/utils/useDebounce/useDebounce.ts
rabelloo: Ah auto-inserted when I open a JSDocs comment.
Nice catch, thanks.

PR #936, opened 2021-08-25
chore: optimising sideEffect for importing CSS

zoeradkani:
## Purpose

CSS can't be sideEffect free as it does not have an export, therefor when importing css into JS you lose it during the build process. 

## Approach
Excluding it from tree shaking.

## Testing

N/A

## Risks
There is no risk, apart from an effect on bundle size

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -5,7 +5,9 @@
   "author": "Onfido",
   "license": "Apache-2.0",
   "type": "module",
-  "sideEffects": false,
+  "sideEffects": [
+    "*.css"
+  ],
   "repository": {
     "type": "git",
     "url": "https://github.com/onfido/castor.git",


rabelloo:
To clarify, this is because Gatsby (and probably webpack) will tree-shake `import '@onfido/castor/dist/castor.css';` out.
https://github.com/gatsbyjs/gatsby/issues/19446

Arguably, the side-effect is introduced by the bundler when it decides that a CSS import requires a DOM element to be created, not the import itself.
But semantics.

I do believe this is a good change overall, improves the lives of Gatsby/webpack users.

rabelloo (approved):

josokinas (approved):

josokinas:
Such change is also needed on https://github.com/onfido/castor-tokens/tree/main/packages/npm package, which Castor is using behind the scenes.

PR #962, opened 2021-09-07
chore(deps): update jest

rabelloo:
## Purpose

Update Jest related dependencies.

## Approach

`npx npm-upgrade` then `yarn`

## Testing

CI

## Risks

None.


.nvmrc
```diff
@@ -1 +1 @@
-14
+14.17
```
josokinas, line 1: That's interesting, so Jest requires a minimum specific minor version of Node? 🤔

Also, worth updating `CONTRIBUTING.md`, it has also mention of Node 14 version.

josokinas (approved):
Looks good, so it wasn't picked up by Dependabot?

rabelloo:
> Looks good, so it wasn't picked up by Dependabot?

It was, but it couldn't automatically resolve the Node version bump.

.nvmrc
rabelloo: Apparently so, or at least lists it as a dependency and `yarn` wants to enforce it.

Will update that file, thanks for the reminder.

PR #966, opened 2021-09-08
build: use esbuild rather than babel

rabelloo:
## Purpose

1. Speed up build (and local development)
2. Avoid infinite rebuild loops in E2E runs by disabling watch mode
3. Migrate Storybook's PostCSS as informed on deprecation

## Approach

1. Replace all `babel-loader` instances in Storybook's webpack config with `esbuild-loader`
2. Check for `NODE_ENV === 'e2e'` and set `watchOptions` to ignore everything
3. Setup `@storybook/addon-toolbars` as instructed on migration guide

## Testing

`yarn start` and preview deploy

## Risks

Will potentially be replaced by the Vite builder in the future, which for Storybook is supposed to be even faster.


.storybook/main.ts
```diff
@@ -3,18 +3,37 @@
...
   ],
-  stories: ['../{docs,packages}/**/*.stories.ts{,x}'],
+
   webpackFinal: (config) => {
+    // minify with esbuild
+    config.optimization = {
+      minimizer: [new ESBuildMinifyPlugin({ target: 'es2019' })],
+    };
+
+    // replace babel with esbuild
```
josokinas, line 27:
```suggestion
    // replace Babel with esbuild
```

.storybook/main.ts
```diff
@@ -23,22 +42,30 @@ module.exports = {
...
       ],
     });
-    config.performance = {
-      hints: false, // https://webpack.js.org/configuration/performance/#performancehints
-    };
+
+    // https://webpack.js.org/configuration/performance/#performancehints
+    config.performance = { hints: false };
+
+    // resolve @onfido/castor imports
```
josokinas, line 49: Maybe better:

```suggestion
    // resolve Castor imports
```

Because `@onfido/castor` seems more like a core package (without React one).

josokinas (approved):
Nice! 👍

PR #1009, opened 2021-09-27
fix: adds missing tooltip and popover exports

davids868:
## Purpose

We want to import `Popover` and `Tooltip` directly from `@onfido/castor-react`

## Approach

re-export the components in the index file.

## Testing

Stories updated with the right imports.

## Risks

N/A


rabelloo (approved):

rabelloo:
Commit message (and PR title) should be a `fix` so that we can release it as `1.5.2`

Also we need to bump bundlewatch's react/prototype to 60kb

PR #1019, opened 2021-09-29
docs(react): improve popover

rabelloo:
## Purpose

Better document how to "open/close" Popover, as well as `onClose` and `target`.

## Approach

Add 2 new stories and change Playground a bit.

## Testing

Preview.

## Risks

- Wording might be confusing
- Code examples might suggest suboptimal solutions
- Story would be better to maintain as MDX


packages/core/src/components/button/button.story.tsx
```diff
@@ -6,7 +6,12 @@ export interface ButtonProps extends BaseProps {
   href?: string;
 }
 
-export const Button = ({ href, kind, variant, ...props }: ButtonProps) =>
+export const Button = ({
+  href,
+  kind = 'action',
+  variant = 'primary',
```
josokinas, line 12: This was actually done on purpose, because for CSS there is no defaults, you *must* set every modifier manually.

And I don't even think this should belong on this PR?

packages/core/src/components/popover/popover.stories.ts
```diff
@@ -46,14 +53,41 @@ export default {
...
 } as Meta<PopoverProps>;
 
-export const Playground: Story<PopoverProps> = (props) =>
+type PlaygroundProps = PopoverProps & { show?: boolean };
+export const Playground: Story<PlaygroundProps> = ({ show, ...props }) =>
+  html('div', {
+    style: { position: 'relative' },
+    children: [
+      Button({ children: 'Target' }),
+      show && Popover(props), //
```
josokinas, line 62: Did you want to comment something here?

josokinas (approved):
You don't need a scope here, I see a change on both core and react packages.

PR #1024, opened 2021-10-01
Fix / popover and tooltip pointer-events

rabelloo:
NOTE: will get merged as two different fixes, no squash, but still one PR because the second commit depends on the first

## Purpose

- Allow Popover to have `pointer-events`. Most content inside Popovers are clickable, like a dropdown menu
- Keep Tooltip visible on its own `:hover` state, as well as its target's

## Approach

- Use `visibility` instead of `pointer-events`
- Add `:hover` rule for Tooltip

## Testing

Preview

## Risks

None.

### DIFF
--- a/packages/core/src/components/popover/popover.scss
+++ b/packages/core/src/components/popover/popover.scss
@@ -9,8 +9,8 @@
 
 @mixin _Popover() {
   .ods-popover-anchor {
-    pointer-events: none;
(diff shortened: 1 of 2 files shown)


josokinas (approved):
Should this be `fix(core)`?

rabelloo:
Hmm probably yea, will amend.

PR #1077, opened 2021-10-15
docs: update castor-tokens migration

rabelloo:
## Purpose

Update `castor-tokens` migration.

## Approach

Update `MIGRATION.md`.

## Risks

We should only merge after [`castor-tokens`' breaking changes](https://github.com/onfido/castor-tokens/pull/245) are merged.

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -2,6 +2,20 @@
 
 ## 1.x to 2.x
 
+### Color tokens (CSS and React)
+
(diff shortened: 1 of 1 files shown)


josokinas:
I think such migration should also be on Castor Tokens repo.

rabelloo:
> I think such migration should also be on Castor Tokens repo.

It'll be part of `CHANGELOG.md` there as soon as we merge.

Do you think we should also add a `MIGRATION.md` there?

PR #1101, opened 2021-10-27
chore: bump node required to lts v16.15

josokinas:
## Purpose

Node LTS is now at v16.15.

## Approach

Require to use Node LTS v16.15.

## Testing

Locally, and CI.

## Risks

N/A


Dockerfile.cypress
```diff
@@ -1,4 +1,4 @@
-FROM cypress/base:14.17.0
+FROM cypress/base:16.13.0
```
rabelloo, line 1: It seems like Cypress' latest Node 16 image is 16.5 unfortunately, so perhaps we target that until it's updated?
https://github.com/cypress-io/cypress-docker-images/tree/master/base

rabelloo: They probably do, but the last update was over a month ago

rabelloo: Updated with latest available, `16.5.0`

rabelloo: `16.13.0` is now available, reverting

rabelloo (approved):

PR #1130, opened 2021-11-15
fix(react): select onChange

rabelloo:
## Purpose

React `<Select/>` is not emitting `onChange` events.
Fix it so it does.

## Approach

Inside `<CustomSelect/>`, when the `select` function is called, emit a synthetic event for 'change' from the native `<select/>` as if the user selected the option directly in it, instead of React modifying its value which is what actually happens.

## Testing

- `onChange={console.log}` listener temporarily added in story to confirm.
- internal Onfido app Signal Builder (where we found the bug) has an e2e test that depends on that event, after the change it passes

## Risks

If anything it reduces risk

### DIFF
--- a/packages/react/src/components/select/select.tsx
+++ b/packages/react/src/components/select/select.tsx
@@ -8,7 +8,7 @@ import { SelectProvider } from './useSelect';
 export type SelectProps =
   | ({ native: true } & BaseProps & NativeSelectProps)
   | ({ native?: false } & BaseProps &
-      Omit<CustomSelectProps, 'open' | 'onOpenChange' | 'onSelectOption'>);
+      Omit<CustomSelectProps, 'open' | 'onOpenChange'>);
 
 /**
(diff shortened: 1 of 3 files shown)


packages/react/src/components/select/select.tsx
```diff
@@ -51,37 +55,24 @@ interface ContentProps extends CustomSelectProps {
...
-    );
+  if (native) return <NativeSelect {...restProps} />;
 
   return (
     <CustomSelect
       {...restProps}
       borderless={borderless}
       open={open}
       onOpenChange={onOpenChange}
-      onSelectOption={(value) => onEmptyChange(!value)}
```
rabelloo, line 84: This should've been the writing on the wall, needing a custom event rather than `onChange`.
Something to watch out for next time

josokinas (approved):
Looks good, though issue in "Unchanged files with check annotations" section.

PR #1140, opened 2021-11-18
fix(react): select empty modifier

rabelloo:
## Purpose

Select `.-empty` modifier (for placeholders) is incorrectly applied when `defaultValue` is used, rather than `value`.

## Approach

The problem is that the useEffect hook that trigger on `value` changes also triggers on first render, just after `useState(empty)`.

In that hook, we should only `setEmpty` if `defaultValue` is not defined.

## Testing

Found in Signal Builder, reproduced and fixed in Storybook

## Risks

None.

### DIFF
--- a/packages/react/src/components/select/select.tsx
+++ b/packages/react/src/components/select/select.tsx
@@ -27,7 +27,9 @@ export const Select = ({
   const [empty, setEmpty] = useState(!(value ?? defaultValue));
   const [open, setOpen] = useState(false);
 
-  useEffect(() => setEmpty(!value), [value]);
+  useEffect(() => {
+    if (defaultValue == null) setEmpty(!value);
+  }, [value]);
 
   return (
     <div
--- a/packages/react/src/components/select/select.tsx
+++ b/packages/react/src/components/select/select.tsx
@@ -28,7 +28,7 @@ export const Select = ({
   const [open, setOpen] = useState(false);
 
   useEffect(() => {
-    if (defaultValue == null) setEmpty(!value);
+    if (value != null) setEmpty(!value);
   }, [value]);
 
   return (


packages/react/src/components/select/select.tsx
```diff
@@ -27,7 +27,9 @@ export const Select = ({
   const [empty, setEmpty] = useState(!(value ?? defaultValue));
   const [open, setOpen] = useState(false);
 
-  useEffect(() => setEmpty(!value), [value]);
+  useEffect(() => {
+    if (defaultValue == null) setEmpty(!value);
```
josokinas, line 31: Shouldn't `defaultValue` be used only once?

Now, when changed alongside `value` - the `setEmpty` will be triggered.

That, is, when `defaultValue` is "a", and actual value is changed from "b" to "" - an empty state will not be applied.

rabelloo: React will complain that you should not use both at the same time, either `defaultValue` or `value` (alongside `onChange`)

rabelloo: An alternative is doing:

```ts
  useStateEffect(() => setEmpty(!(value ?? defaultValue)), [value]);
```

Which reads almost the same, supports the use case you mention, but arguably is opposite to React's API.

josokinas: I'm thinking, should you set a state of a default value, and use that instead? So it would only bet read from props once, and updating it would not change the behaviour?

rabelloo: That's what we had before, but there are scenarios in which we do want to "revert" to a "placeholder" state.
The simplest one is a controlled Select with a non-disabled, non-hidden "placeholder" option.

```tsx
const [value, setValue] = useState('');

<Select value={value} onChange={(ev) => setValue(ev.target.value)}>
  <Option value="">Select an option...</Option>
  <Option value="foo">foo</Option>
  <Option value="bar">bar</Option>
</Select>
```

If you select `foo`, then `''` again, it should reapply the `.-empty` modifier class.

josokinas: Sure, that's fine, but I still think that this:

```js
if (defaultValue == null) setEmpty(!value);
```

should actually be:

```js
if (initialDefaultValue == null) setEmpty(!value);
```

So that changing `initialValue` would not take any affect on the function outcome.

But your call, can leave like this also.

josokinas (approved):

packages/react/src/components/select/select.tsx
rabelloo: I see. We'd need `const [initialDefaultValue] = useState(defaultValue)`.

Perhaps we should go for the most dynamic then?

```ts
useEffect(() => setEmpty(!(value ?? defaultValue)), [value]);
```

Theoretically it is possible and valid to switch between controlled and uncontrolled, like:

```tsx
const [value, setValue] = useState<string>();

useEffect(() => void fetch('foo').then(setValue), []);

<Select
  defaultValue={value == null ? undefined : ''}
  value={value}
  onChange={(ev) => setValue(ev.target.value)}
>
  <Option value="">Select an option...</Option>
  <Option value="foo">foo</Option>
  <Option value="bar">bar</Option>
</Select>
```

Although very weird, sure, but still, we could support it with simple enough code?
What do you think?

josokinas: This looks very odd.

We can live with original solution probably, then see if there is a problem.

If not broken - not worth fixing, right 😏

rabelloo: Right. So are you saying:

1. Only read `value` and `defaultValue` once, set `empty` accordingly, and remove the `useEffect` hook
2. Set `empty` when either changes, but in a consistent manner, with:
```
useEffect(() => setEmpty(!(value ?? defaultValue)), [value]);
```

josokinas: I would prefer option 1, but reading only `defaultValue` once - reading `value` always is fine.

rabelloo: I found out that the rabbit hole goes deeper.
I'll open another Pull Request in a bit.

For this one, we can limit it to when `value != null`, which is in the latest commit.
That fulfils both our requirements, and mirrors the native Select element and React's API.

josokinas (approved):

PR #1150, opened 2021-11-24
fix(react): select should respond to value changes

rabelloo:
## Purpose

React's `<Select />` does not update when its `value` changes, but it should.

## Approach

Introduce `useEffect` hooks that set internal value when it changes.
Also change `initialize` a bit so that the correct option JSX can be found with its value as key.

## Testing

Storybook with `value` property defined and changed with, for example, hooks and buttons

## Risks

Slightly increases `Select`'s bundle size

### DIFF
--- a/packages/react/src/components/select/custom/custom-select.tsx
+++ b/packages/react/src/components/select/custom/custom-select.tsx
@@ -2,6 +2,7 @@ import { c, classy, m, PopoverProps, SelectProps } from '@onfido/castor';
 import React, {
   ReactNode,
   SyntheticEvent,
+  useEffect,
   useMemo,
   useRef,
   useState,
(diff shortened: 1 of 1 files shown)


packages/react/src/components/select/custom/custom-select.tsx
```diff
@@ -31,39 +32,46 @@ export function CustomSelect({
...
+  value,
   ...restProps
 }: CustomSelectProps) {
   const selectRef = useRef<HTMLSelectElement>(null);
-  const [selectedOption, setSelectedOption] = useState<ReactNode>();
-  const [value, setValue] = useState(restProps.value ?? defaultValue);
+  const options = useRef(new Map<typeof value, ReactNode>());
+  const [currentValue, setCurrentValue] = useState<typeof value>();
+
+  // default to first option after first render
```
josokinas, line 42: I think this is more accurate:

```suggestion
  // default to first option on first render
```

josokinas (approved):

PR #1172, opened 2021-11-30
chore(react): remove unused selectedOption

rabelloo:
## Purpose

Some clean up.

## Risks

None, internal API


josokinas (approved):
IMO it's not a `chore` but probably `style`? As it is touching `src`.

rabelloo:
Good point, will rename

PR #1174, opened 2021-11-30
fix(react): ensure output's height

rabelloo:
## Purpose

Make sure Select has consistent height even if no output is rendered, when there are no options in it.

## Approach

Always print a no-break-space at the end, ensuring the line-height will be applied by the browser.

## Testing

Storybook, provide no children for a Select component.

## Risks

None, does not affect its value or visuals.

### DIFF
--- a/packages/react/src/components/select/custom/custom-select.tsx
+++ b/packages/react/src/components/select/custom/custom-select.tsx
@@ -115,7 +115,10 @@ export function CustomSelect({
         {!currentValue || <option hidden value={currentValue} />}
       </NativeSelect>
 
-      <output className={classy(c('select-output'))}>{selectedOption}</output>
+      <output className={classy(c('select-output'))}>
+        {/* no-break space guarantees element height */}
+        {selectedOption}&nbsp;
+      </output>
 
       {isOpen && (
         <Popover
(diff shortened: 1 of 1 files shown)


packages/react/src/components/select/custom/custom-select.tsx
```diff
@@ -107,7 +107,10 @@ export function CustomSelect({
         {!currentValue || <option hidden value={currentValue} />}
       </NativeSelect>
 
-      <output className={classy(c('select-output'))}>{selectedOption}</output>
+      <output className={classy(c('select-output'))}>
+        {/* no-break space guarantees element height */}
+        {selectedOption}&nbsp;
```
josokinas, line 112: Maybe this is more noticable?

```suggestion
        {selectedOption}&nbsp;{/* no-break space guarantees element height */}
```

josokinas (approved):

packages/react/src/components/select/custom/custom-select.tsx
rabelloo: Definitely, we can also put it in a new line for an even clearer effect

```suggestion
        {selectedOption}
        &nbsp; {/* no-break space guarantees element height */}
```

PR #1181, opened 2021-11-30
docs: migrate to CSF3

rabelloo:
## Purpose

Migrate from CSF2 to [CSF3](https://storybook.js.org/blog/component-story-format-3-0/)

## Approach

Follow [Storybook's guide](https://storybook.js.org/blog/component-story-format-3-0/)

## Testing

Storybook should continue to display the same thing

## Risks

New format is a bit more verbose for a small amount of use cases, but better for most and also enables [interaction testing](https://storybook.js.org/blog/interaction-testing-sneak-peek/)

### DIFF
--- a/docs/helpers/matrix/htmlMatrix.ts
+++ b/docs/helpers/matrix/htmlMatrix.ts
@@ -1,3 +1,5 @@
+import type { FC } from 'react';
+import { omit } from '../docs';
 import { Story } from '../story';
 import { getEntries } from './getEntries';
 import { Entries } from './types';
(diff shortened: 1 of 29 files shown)


josokinas (approved):
Blindly approving this - as long as all still functions 😏

rabelloo:
I'd appreciate it if you could take a look at one change example, such as `button.stories.tsx`, just to see what's the new API like.

In my opinion it makes the regular matrices we have for individual properties much cleaner (with our helper), but it makes the "all combinations" stories a bit more verbose

josokinas:
All good, I don't mind either - it's only a matter of getting used to.

BTW, Docs tab is not working, is this expected?

rabelloo:
Yes, it requires the fix in https://github.com/onfido/castor/pull/1180

PR #1203, opened 2021-12-03
docs: color examples

rabelloo:
## Purpose

Add examples to color documentation.

## Approach

Display all colours dynamically depending on theme selected.

## Testing

Storybook > Castor > Color > Examples

## Risks

- ~~Non-component stories are currently out of order, but that regression was introduced some time ago.~~
  ~~Will open another PR to fix that~~
  Actually that only happens locally, less priority then

- Not the best visualisation, but sets a minimum that we can improve over time.
  Documentation as a whole will continue to be worked on in the near future

### DIFF
--- a/docs/_core/colors/colors.scss
+++ b/docs/_core/colors/colors.scss
@@ -1,0 +1,55 @@
+@use '~@onfido/castor';
+
+.cards {
+  display: flex;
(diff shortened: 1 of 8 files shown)


josokinas (approved):
👏

PR #1204, opened 2021-12-03
docs: syntax highlight for sass

rabelloo:
## Purpose

Fix Sass/CSS syntax highlight in stories.

## Approach

Use `css` rather than `scss` or `sass` since only the first is supported.

## Testing

Storybook -> Castor stories

## Risks

None


josokinas (approved):
A bit odd to have Sass examples with "css" highlighting, but as long as it works ™️

PR #1337, opened 2022-01-11
ci: remove Docker from E2E tests

rabelloo:
## Purpose

E2E tests are not only slow to run (2 layer virtualisation: GitHub's virtual environment + Docker) but quite unreliable.
Flakiness, in specific, prevents a lot of PRs to be merged even though they are otherwise completely fine.
Then, we have to rerun those failing jobs but they take ~30 minutes to run.
So, removing Docker out of the equation to make it much faster again.

## Approach

Instead of guaranteeing the same environment with a Docker image, we make GitHub actions the _de-facto_ environment once again.
Except this time we make updating the images when the diffing fails (intentional visual changes) much easier with a script that downloads the artifact from GitHub's API, using its CLI.

Tried using [act](https://github.com/nektos/act) to execute workflows locally but no luck.
Might invest more effort in the future if we find it necessary.

## Testing

E2E tests should finally pass in this PR, all others are failing.

## Risks

Not being able to generate "baseline screenshots" locally, and always depend on GitHub.
However GitHub availability is not something we need to worry about, if Actions can't be run or artifacts can't be downloaded, most likely the whole platform is down and we can't do anything anyway.
And there are other alternatives like the aforementioned `act` which we can further explore.

Also the branch must be pushed to GitHub for baseline screenshots to be taken, but also not a big worry since PRs are the very place we want to check for visual regressions, and at that point the branch must be pushed.


package.json
```diff
@@ -10,38 +10,37 @@
...
     "poste2e": "nyc report -r text -r text-summary",
-    "e2e:run": "CYPRESS_baseUrl=http://host.docker.internal:6006 docker-compose up --abort-on-container-exit --exit-code-from cypress",
+    "e2e:run": "wait-on http://localhost:6006 && yarn cypress run -c video=true",
     "e2e:serve": "NODE_ENV=e2e yarn start --ci --quiet",
-    "snapshot": "CYPRESS_updateSnapshots=true yarn e2e",
-    "inspect": "concurrently yarn:inspect:*",
-    "preinspect:bundle-size": "yarn build && yarn prototype",
-    "inspect:bundle-size": "bundlewatch --config .bundlewatchrc.json",
+    "screenshot": "CYPRESS_updateSnapshots=true yarn e2e",
+    "baseline": "FOLDER=e2e/.screenshots; shx rm -rf $FOLDER &&  gh run download -n screenshots -D $FOLDER $(gh run list -w test | grep $(git branch --show-current) | sed 's/.*\\([0-9]\\{10\\}\\).*/\\1/g')",
```
rabelloo, line 41: Hate how bad this script reads but only alternatives are either multiple `package.json` scripts, or a custom file with either a Node or Bash script, of which I like neither :/

.gitignore
```diff
@@ -26,6 +26,6 @@ package-lock.json
 coverage
 cypress
 dist
-public
+/public
```
josokinas, line 29: Will this ignore `public` from examples now?

CONTRIBUTING.md
```diff
@@ -1,32 +1,32 @@
 # Contributing
 
-When contributing to this repository, please read through and get familiarised with the processes of making a change to this package.
+Please read through and get familiarised with the processes for introducing a
+change in Castor.
```
josokinas, line 4: Why sentences are now split into lines? IMO having one single line is better for different screen sizes.

CONTRIBUTING.md
```diff
@@ -1,32 +1,32 @@
...
+We recommend using [nvm](https://github.com/nvm-sh/nvm) to manage multiple Node
+versions.
 
-To install and switch to required version run:
+To install and switch to version specified in `.nvmrc` run:
 
     nvm install && nvm use
 
-Finally install Yarn via Homebrew:
+You'll also need [yarn](https://classic.yarnpkg.com/en/docs/install) which you
```
josokinas, line 22: `Yarn` spelling:

```suggestion
You'll also need [Tarn](https://classic.yarnpkg.com/en/docs/install) which you
```

CONTRIBUTING.md
```diff
@@ -1,32 +1,32 @@
...
+can install with `npm`:
 
-    brew install yarn
-
-Then remove Node that was installed alongside Yarn:
-
-    brew uninstall node --ignore-dependencies
+    npm i -g yarn
```
josokinas, line 25: Isn't this for Yarn 2? Which does not work for us.

josokinas (approved):

CONTRIBUTING.md
rabelloo: You mean uppercase right?

```suggestion
You'll also need [Yarn](https://classic.yarnpkg.com/en/docs/install) which you
```

CONTRIBUTING.md
rabelloo: No Yarn 2 wants us to use corepack https://yarnpkg.com/getting-started/install

Installing with `npm -g` is for v1 https://classic.yarnpkg.com/en/docs/install#mac-stable

.gitignore
rabelloo: I believe those are all `dist` but if they did create such a folder it would not be ignored

The goal here is to ignore `./public` but not `.storybook/public`

PR #1373, opened 2022-01-20
fix: button hover on iOS

nulrich:
## Purpose

On iOS Safari, if you tap on an element, the hover state "sticks" and any element displayed at the same place will be in hover state.

Here's a sample to reproduce:
```
import React from "react";
import { Button } from "@onfido/castor-react";
import "@onfido/castor/dist/castor.css";
import "@onfido/castor/dist/themes/day.css";

export default function App() {
  const [page, setPage] = React.useState(1);

  const PageOne = () => <Button onClick={() => setPage(2)}>PageOne</Button>;
  const PageTwo = () => <Button onClick={() => setPage(1)}>PageTwo</Button>;

  return page === 1 ? <PageOne /> : <PageTwo />;
}

```

## Approach

Wrapping every `:hover` rule in `@media (hover: hover)` to prevent mobile device to apply the rule.

## Testing

Should pass CI since it doesn't apply to desktop browser.

## Risks

Adding the media query trigger the `max-nesting-depth`, so the rule has been disabled locally.

### DIFF
--- a/packages/core/src/components/button/button.scss
+++ b/packages/core/src/components/button/button.scss
@@ -37,9 +37,13 @@
       background-color: helpers.color('background-action');
       color: helpers.color('content-on-action');
 
-      &:hover {
-        background-color: helpers.color('background-action-hover');
+      /* stylelint-disable max-nesting-depth */
+      @media (hover: hover) {
+        &:hover {
+          background-color: helpers.color('background-action-hover');
+        }
       }
+      /* stylelint-enable max-nesting-depth */
 
       &:active {
         background-color: helpers.color('background-action-active');
@@ -49,19 +53,27 @@
     &.-action--secondary {
       border-color: helpers.color('border-action');
 
-      &:hover,
-      &:active {
-        border-color: helpers.color('border-action-hover');
+      /* stylelint-disable max-nesting-depth */
+      @media (hover: hover) {
+        &:hover,
+        &:active {
+          border-color: helpers.color('border-action-hover');
+        }
       }
+      /* stylelint-enable max-nesting-depth */
     }
 
     &.-action--secondary,
     &.-action--tertiary {
       color: helpers.color('content-action');
 
-      &:hover {
-        background-color: helpers.color('background-action-subtle-hover');
+      /* stylelint-disable max-nesting-depth */
+      @media (hover: hover) {
+        &:hover {
+          background-color: helpers.color('background-action-subtle-hover');
+        }
       }
+      /* stylelint-enable max-nesting-depth */
 
       &:active {
         background-color: helpers.color('background-action-subtle');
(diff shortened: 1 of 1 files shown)


packages/core/src/components/button/button.scss
```diff
@@ -49,19 +53,27 @@
     &.-action--secondary {
       border-color: helpers.color('border-action');
 
-      &:hover,
-      &:active {
-        border-color: helpers.color('border-action-hover');
+      /* stylelint-disable max-nesting-depth */
+      @media (hover: hover) {
+        &:hover,
+        &:active {
```
rabelloo, line 59: I don't think we should wrap `:active` here too, we probably still want that style to apply in touch devices that don't support `hover`, i.e. iOS

packages/core/src/components/button/button.scss
```diff
@@ -37,9 +37,13 @@
       background-color: helpers.color('background-action');
       color: helpers.color('content-on-action');
 
-      &:hover {
-        background-color: helpers.color('background-action-hover');
+      /* stylelint-disable max-nesting-depth */
```
rabelloo, line 40: [minor] didn't `stylelint-disable-next-line` work? It would cut these comments in half

nulrich: If I use `stylelint-disable-next-line` I need to write :

```
      /* stylelint-disable-next-line max-nesting-depth */
      @media (hover: hover) {
        /* stylelint-disable-next-line max-nesting-depth */
        &:hover {
          background-color: helpers.color('background-negative-subtle-hover');
        }
      }
```

So I found that using disable/enable for more clear

rabelloo: Ah because we ignore pseudo-classes, so now it's 3 depth...
Alright, looks better the way it is then, thanks for clarifying

packages/core/src/components/button/button.scss
```diff
@@ -53,14 +53,10 @@
     &.-action--secondary {
       border-color: helpers.color('border-action');
 
-      /* stylelint-disable max-nesting-depth */
-      @media (hover: hover) {
-        &:hover,
-        &:active {
-          border-color: helpers.color('border-action-hover');
-        }
+      &:hover,
```
rabelloo, line 56: We still want to wrap `:hover` with the media query though, right?
So maybe we need two rules now

```scss
@media (hover:hover) {
    &:hover {
      border-color: helpers.color('border-action-hover');
    }
}

&:active {
  border-color: helpers.color('border-action-hover');
}
```

rabelloo (approved):

rabelloo:
Thanks for your contribution!
I'll release as soon as possible, hopefully today but if not, next Monday

PR #1566, opened 2022-04-06
docs: fix some component names

rabelloo:
## Purpose

Some component names are not being correctly inferred by Storybook tools.
It is unclear why, as it was working in previous versions.

## Approach

Hard-code problematic components' names in Storybook only.
Ideally we can simply remove this after tools work once again.

## Testing

Check code snippet for all components (e.g. Spinner, Radio, Checkbox) before and after change

## Risks

Temporary code tends to live longer than intended

### DIFF
--- a/.storybook/custom/fixNames.ts
+++ b/.storybook/custom/fixNames.ts
@@ -1,0 +1,25 @@
+import {
+  Checkbox,
+  Fieldset,
+  Form,
+  Input,
+  Progress,
+  Radio,
+  Spinner,
+  Textarea,
+} from '@onfido/castor-react';
+import { FC } from 'react';
+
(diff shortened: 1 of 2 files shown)


josokinas (approved):
Wouldn't it be better to set `displayName` on each component's file?

BTW, maybe it's this that caused it - https://github.com/onfido/castor/pull/869?

rabelloo:
that would increase bundle size, and not provide much value

I can see it being important for React browser extensions, which probably rely on those, but I haven't tested yet whether or not they present the same issue as Storybook

it's very likely that was the cause, although it's weird it still works for some components but not others

rabelloo:
I'll merge as is for a fix to our docs and we can change it later if necessary

Issue #64, opened 2020-12-08
ComboBox


coreyhume:
Implement the ComboBox component.

PR #1727, opened 2022-05-25
feat(react): combobox

rabelloo:
## Purpose

Implement [Combobox](https://github.com/onfido/castor/issues/64) component

## Approach

Extract `OptionList` out of `Select` for reusability, then create `Combobox` that reuses the former.

Mostly composable, small amount of repetition, the biggest difference is event bindings and what each does, not worth a propscalypse on `CustomSelect` for that.

## Testing

Storybook.

## Risks

1. Arguable breaking change in `Select`.
I'd say it's not breaking because it's internals.

2. Non-React component is extremely complex and convoluted, not sure if even worth an HTML + CSS story.
Web component would be nice but we still don't do that


Subcide:
I did some more testing, I've found the following issues, and tried to detail more of the accessibility approach.

- When the combobox is focussed to via keyboard, the options list shouldn’t display until the up or down cursor keys are pressed, or the user has started typing a word.  This is in-keeping with expected accessible patterns as described in the aria docs, as well as other implementations I've seen.  I know you were pushing back on this a bit Andre, but I think we need it for it to be properly usable and accessible.
- On the vercel instance I’m testing on, Select appears to be broken (placeholder text or selected option not displaying).  Need to investigate if this is a breaking change due to combobox changes, or something unrelated.

Accessibility:
(Tested on Safari with Voiceover on Mac.)
Combobox needs aria attributes specified as explained here: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/combobox_role#specifications
It's hard for me to add roles and aria stuff to see how different things affect what's read out, so I'd suggest we start with the above recommendations, and iterate/improve from there.

TheComfyChair (approved):

zoeradkani (approved):

